PR #1, opened 2020-07-25
Improve docgen type information

cmb69:
We add support for return types and union types, and properly retrieve
the types via the `::getType()` and `::getReturnType()` methods,
repectively.

### DIFF
--- a/scripts/docgen/docgen.php
+++ b/scripts/docgen/docgen.php
@@ -1,7 +1,7 @@
 <?php
 /*  
   +----------------------------------------------------------------------+
-  | PHP Version 5                                                        |
+  | PHP Version 7                                                        |
   +----------------------------------------------------------------------+
   | Copyright (c) 1997-2011 The PHP Group                                |
   +----------------------------------------------------------------------+
@@ -17,7 +17,7 @@
   |             Philip Olson <philip@php.net>                            |
   +----------------------------------------------------------------------+
  
-  $Id$
+  $Id: docgen.php 336015 2015-02-27 20:31:12Z aharvey $
 */
 
 if (!(extension_loaded('reflection') && extension_loaded('pcre'))) {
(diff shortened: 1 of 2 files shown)


scripts/docgen/docgen.php
```diff
@@ -17,7 +17,7 @@
   |             Philip Olson <philip@php.net>                            |
   +----------------------------------------------------------------------+
  
-  $Id$
+  $Id: docgen.php 336015 2015-02-27 20:31:12Z aharvey $
```
carusogabriel, line 20: We still need this in the docs, right? :(

scripts/docgen/docgen.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /*  
   +----------------------------------------------------------------------+
-  | PHP Version 5                                                        |
+  | PHP Version 7                                                        |
```
carusogabriel, line 4: We've removed this in the core, if you want to drop it here as well: https://github.com/php/php-src/pull/4732

scripts/docgen/docgen.php
cmb69: I don't think that this is still needed (we need the IDs of the documentation XML files for the translations, though).

scripts/docgen/docgen.php
cmb69: Yep, doesn't make much sense to have it here, but I assume there are several other files where it could be removed; might better be done as separate commit.

cmb69:
If there are no objections, I'll merge this in a week.

scripts/docgen/docgen.php
```diff
@@ -202,6 +202,26 @@ function get_type_by_string($str) { /* {{{ */
 }
 /* }}} */
 
+/** @return string|null */
+function get_type_as_string(ReflectionType $type = null) { /* {{{ */
```
salathe, line 206: Why not just `return (string) $type` for all non-null inputs?

scripts/docgen/docgen.php
```diff
@@ -1085,7 +1116,7 @@ function get_default_role ($role, $funcname, $default) {
 	switch ($opt) {
 		case 'v':
 		case 'version':
-			printf("%s\n", '$Revision$');
+			printf("%s\n", '$Revision: 336015 $');
```
salathe, line 1119: Unwanted change. Please leave it as it was.

scripts/docgen/docgen.php
cmb69: Because [ReflectionType::__toString()](https://www.php.net/manual/en/reflectiontype.tostring.php) is deprecated as of PHP 7.1.0.

scripts/docgen/docgen.php
cmb69: Thanks, done.

scripts/docgen/docgen.php
cmb69: Thanks, done.

scripts/docgen/docgen.php
cmb69: Thanks, done.

scripts/docgen/docgen.php
salathe: Ack.

cmb69:
Thanks! Applied as http://svn.php.net/viewvc?view=revision&revision=350350.

PR #3, opened 2020-11-03
Actually add PHP 8.0 migration guide to the manual

cmb69:
This requires https://github.com/php/doc-en/pull/170.


cmb69:
Applied as http://svn.php.net/viewvc?view=revision&revision=351501.

PR #7, opened 2020-12-30
chm generate batch script migrated from svn to git too!

mumumu:
Finally announced abount migrating from svn to git! 
We need to migrate not only docbook related stuff, but also chm generate batch!

- changed version control command, from svn to git
- deleted inactive korean setting

Note: I tried to push this changeset to git.php.net directly, but could'nt do that because of insufficient Karma.
So, I send these changes here.


heiglandreas:
Thanks for taking care of this!

PR #9, opened 2020-12-31
Enhancement: Build documentation on GitHub Actions

localheinz:
This PR

* [x] builds the documentation for (hopefully) all languages on GitHub Actions
* [x] works around unnecessary modification of the root directory on GitHub Actions

💁‍♂️ You can see this in action at https://github.com/localheinz/doc-base/pull/1.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -1,0 +1,54 @@
+# https://docs.github.com/en/actions
+
+name: "Integrate"
+
+on:
+  pull_request: null
+  push:
+    branches:
+      - "master"
+
+jobs:
+  build:
+    name: "Build"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        language:
+          - "de"
+          - "en"
+          - "es"
+          - "fr"
+          - "it"
+          - "ja"
+          - "pl"
+          - "pt_br"
+          - "ro"
+          - "ru"
+          - "tr"
+          - "uk"
+          - "zh"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          path: "doc-base"
+
+      - name: "Checkout php/doc-${{ matrix.language }}"
+        uses: "actions/checkout@v2"
+        with:
+          path: "${{ matrix.language }}"
+          repository: "php/doc-${{ matrix.language }}"
+
+      - name: "Checkout php/doc-en as fallback"
+        if: "matrix.language != 'en'"
+        uses: "actions/checkout@v2"
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,63 @@
...
+          - "fr"
+          - "it"
+          - "ja"
+          - "pl"
+          - "pt_br"
+          - "ro"
+          - "ru"
+          - "tr"
+          - "uk"
+          - "zh"
```
localheinz, line 35: Are we missing any languages here?

Does it even make sense to build all languages here as well, or should we only use two languages, perhaps `de`, and `en`, to test building `en`, and also, to test the fallback to `en` when building a language other than `en`?

cmb69: Looks like these are all languages for which we have Git repos so far. Not sure about IT, PL and UK repos; these are inactive for the online docs.

I'm not sure whether doc-base changes need to be tested at all, since these are relatively rare, and issues should be caught pretty quickly when testing the languages. OTHO, doesn't hurt to have a full setup here as well.

heiglandreas: each language that is part of the ACTIVE_LANGUAGES array has a repo. Whether we want to deactivate them is a separate question IMO.

williamdes (approved):

localheinz:
@cmb @heiglandreas @williamdes 

Let me know what you think, would be nice to move forward here!

configure.php
```diff
@@ -234,7 +234,15 @@ function print_xml_errors($details = true) {
...
-if (basename($rootdir) == 'doc-base') {
+
+/**
+ * When checking out this repository on GitHub Actions, the workspace  directory is "/home/runner/work/doc-base/doc-base".
+ *
+ * To avoid applying dirname() here, we check if we are running on GitHub Actions.
+ *
+ * @see https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
+ */
+if (getenv('GITHUB_ACTIONS') !== 'true' && basename($rootdir) == 'doc-base') {
```
williamdes, line 245:
```suggestion
if (getenv('GITHUB_ACTIONS') !== 'true' && basename($rootdir) === 'doc-base') {
```

williamdes (approved):
This looks good, It would be cool to have some internal PHP feedback

configure.php
localheinz: 👍

cmb69:
LGTM, thank you! (and please don't ping Chris Brannon; my GH handle is @cmb69 :)

localheinz:
Thank you, @cmb69!

PR #11, opened 2021-01-03
Fix paths in the Quick Reference

cmb69:
We clarify that one has to `cd` into the desired directory, and adjust
the script paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,17 @@ and resources used in all languages.
 ### Edits
 
 * Make the change.  Use spaces not tabs.  Be sure to carefully watch your whitespace!
+* cd into the desired clone directory, e.g.
+  ```bash
+  cd en
+  ```
 * Look at your unified diff, make sure it looks right and that whitespace changes aren't mixed in:
   ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,9 +86,13 @@ and concerns.
 * Edit.  Be sure no leftover text exists.  No tabs either.
 * If you are translating, remember to add the full Git commit hash of the English
   file that you are translating from, to the file's `EN-Revision` comment.
+* cd into the desired clone directory, e.g.
+  ```bash
+  cd en
+  ```
```
Girgias, line 92: Might make sense to say this name is hardcoded? As when you git clone you'd get ``doc-en`` as the folder name, so or a symlink is necessary or ``git clone link-to-doc-en-repo.git en`` should be done while cloning doc-en

cmb69: When cloning from git.php.net (like described further up the page), the default directory name is `en`.

Girgias: Ah, didn't know it was referencing git.php.net directly.

Girgias (approved):

PR #12, opened 2021-01-07
Update global.ent

wenz:
Replace retired IMAP c-library link with copy on GitHub


Girgias (approved):
Makes sense as the current link is a 404

cmb69:
Yes. but should we link to that repository?

Girgias:
It doesn't seem like the university of Washington has an "official" mirror any more, so this is the best next thing IMHO. And as it we're still keeping the IMAP extension around being able to find the sources somehow, albeit not by the original authors still seems valuable to me.

PR #15, opened 2021-01-12
Remove register globals and magic quote pages

Girgias:
Those have been removed before PHP 7


cmb69:
Sorry, this is not going to happen – because it already did. :)

PR #23, opened 2021-04-24
Provides easy access to diffs from GitHub, local and git.php.net

alfsb:
Change the modified file section to provide:

- Local git repository diff command (always work)
- Individual file diff from GitHub web interface (may not work with very old commits)
- Individual file diff from git.php.net, text and html (may not work with very recent commits)

An alternative is to create a "tig.php.net": a mirror from GitHub of canonical repositories managed there, with gitweb installed, to provide a reliable, not external, source of text and html diffs for editors and translators.

### DIFF
--- a/scripts/revcheck.php
+++ b/scripts/revcheck.php
@@ -621,7 +621,7 @@ HTML;
         $d1 = "https://github.com/php/doc-en/compare/{$tr->hash}..{$en->hash}#diff-{$kh}";
         // Local git diff  -- Always work
         $d2 = "(cd en; git diff {$tr->hash} {$en->hash} -- $key)";
-        $d2 = htmlspecialchars( $d2 );
+        $d2 = htmlspecialchars( $d2 , ENT_QUOTES );
         // git.php.net     -- May not work with very recent commits
         $d3 = "https://git.php.net/?p=doc/en.git;a=blobdiff_plain;f=$key;hb={$en->hash};hpb={$tr->hash};"; // text
         $d4 = "https://git.php.net/?p=doc/en.git;a=blobdiff;f=$key;hb={$en->hash};hpb={$tr->hash};";       // html
--- a/scripts/revcheck.php
+++ b/scripts/revcheck.php
@@ -616,9 +616,20 @@ HTML;
             case FileStatusEnum::TranslatedCritial: $bg = 'bgred'   ; break;
             default:                                $bg = 'bggray'  ; break;
         }
-        $d1 = "https://git.php.net/?p=doc/en.git;a=blobdiff_plain;f=$key;hb={$en->hash};hpb={$tr->hash};"; // text
-        $d2 = "https://git.php.net/?p=doc/en.git;a=blobdiff;f=$key;hb={$en->hash};hpb={$tr->hash};";       // html
-        $nm = "<a href='$d1'>{$en->name}</a> <a href='$d2'>(h)</a>";
+        // GitHub web diff -- May not work with very old commits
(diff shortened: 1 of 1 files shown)


scripts/revcheck.php
```diff
@@ -616,9 +616,20 @@ function print_html_files( $enFiles , $trFiles , $lang )
...
         }
-        $d1 = "https://git.php.net/?p=doc/en.git;a=blobdiff_plain;f=$key;hb={$en->hash};hpb={$tr->hash};"; // text
-        $d2 = "https://git.php.net/?p=doc/en.git;a=blobdiff;f=$key;hb={$en->hash};hpb={$tr->hash};";       // html
-        $nm = "<a href='$d1'>{$en->name}</a> <a href='$d2'>(h)</a>";
+        // GitHub web diff -- May not work with very old commits
+        $kh = hash( 'sha256' , $key );
+        $d1 = "https://github.com/php/doc-en/compare/{$tr->hash}..{$en->hash}#diff-{$kh}";
+        // Local git diff  -- Always work
+        $d2 = "(cd en; git diff {$tr->hash} {$en->hash} -- $key)";
+        $d2 = htmlspecialchars( $d2 );
```
TysonAndre, line 624: htmlspecialchars is the wrong method to use with `data-clipboard-text='{$d2}'` - I assume the use of single quotes in the original code is just convenience over `\"`

```
php > echo htmlspecialchars("'");
'
php > echo htmlspecialchars("\"");
&quot;
```

kamil-tekiela:
Why did you not name the variables properly? What's the point in adding comments if you could name the variables? Is there some rule that prevents us from using more than 2 characters for variable names?

scripts/revcheck.php
alfsb: Thanks, and good catch. Added ENT_QUOTES ensure all quotes are always escaped.

alfsb:
> What's the point in adding comments if you could name the variables?

The comments are about the variables contents, not the variables names. It's only a hint for future -- yet unlikely -- debuggers of this sad situation.

cmb69:
What is the point of having diffs with git.php.net, since that is no longer used?

nilgun:
An option for old files,

github.com result:
 https://github.com/php/doc-en/compare/96c9d88bad9a7d7d44bfb7f26c226df7ee9ddf26..f85d380904ca6cfcc5d71d5efe44d6a4a673e2b1#diff-7d8771e20183d6a2a27f3abd85c79eea92a17991e58554fc05a6892fd0279557

git.php.net result:
https://git.php.net/?p=doc/en.git;a=blobdiff;f=reference/image/functions/imageline.xml;hb=f85d380904ca6cfcc5d71d5efe44d6a4a673e2b1;hpb=96c9d88bad9a7d7d44bfb7f26c226df7ee9ddf26;

cmb69:
Okay, I see, but this is of limited usefulness, because **git.php.net is no longer used for weeks**, and it is unlikely that it'll be revived.

alfsb:
On Mon, Apr 26, 2021 at 10:23 AM Christoph M. Becker <
***@***.***> wrote:

> Okay, I see, but this is of limited usefulness, because *git.php.net
> <http://git.php.net> is no longer used for weeks*, and it is unlikely
> that it'll be revived.
>
Well, it's about despair that revcheck.php still points to git.php.net, so
used or not, revived or not, it may keep pointing there while git.php.net
keeps providing some result.

André

cmb69:
Okay, I had a closer look, and this is a viable workaround for now, since GH apparently doesn't support single file comparisons. Thanks for your work!

PR #25, opened 2021-05-14
[RFC] Updated revcheck

saundefined:
- Not all files should be translated, added exceptions that were applied in the [edit.php.net](https://github.com/php/web-doc-editor/blob/master/php/RepositoryManager.php#L1685). 
But I'm not sure that these rules are still actual 🤔 

- In the not translated section added the link to GitHub, where is English content for this file.

- git.php.net link on certain file revision replaced with GitHub link.

### DIFF
--- a/scripts/revcheck.php
+++ b/scripts/revcheck.php
@@ -672,7 +672,7 @@ HTML;
     print <<<HTML
 
  <tr class=bggray>
-  <td class="c">$en->name</td>
+  <td class="c"><a href="https://github.com/php/doc-en/blob/{$en->hash}/$key">$en->name</a></td>
   <td class="c">$en->hash</td>
   <td class="c">$size</td>
  </tr>
--- a/scripts/revcheck.php
+++ b/scripts/revcheck.php
@@ -204,9 +204,6 @@ function populateFileTreeRecurse( $lang , $path , & $output )
                 'DO_NOT_TRANSLATE',
                 'rsusi.txt',
                 'missing-ids.xml',
-                'license.xml',
-                'translation.xml',
-                'versions.xml',
             ];
             
             $ignoredDirectories = [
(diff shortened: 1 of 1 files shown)


scripts/revcheck.php
```diff
@@ -190,6 +190,39 @@ function populateFileTreeRecurse( $lang , $path , & $output )
...
+            $ignoredDirectories = [
+                'chmonly',
+                'internals',
+                'internals2',
+            ];
+            
+            if(
+                in_array($trimPath, $ignoredDirectories, true) 
+                || in_array($filename, $ignoredFileNames, true)
+                || (strpos($filename, 'entities.') === 0)
```
kamil-tekiela, line 218: Are we not running PHP 8 yet? 
Also, the last two `in_array` are missing the strict flag

saundefined: Thanks! Added strict flag, but I'm not sure about PHP 8.

cmb69: This script is supposed to be run locally, AFAIK. Raising requirements to PHP 8.0 might be a bit early; at least that should be discussed on the docs ML.

scripts/revcheck.php
```diff
@@ -190,6 +190,39 @@ function populateFileTreeRecurse( $lang , $path , & $output )
...
+                'README',
+                'DO_NOT_TRANSLATE',
+                'rsusi.txt',
+                'missing-ids.xml',
+            ];
+            
+            $ignoredDirectories = [
+                'chmonly',
+                'internals',
+                'internals2',
```
Girgias, line 212: Didn't we remove those docs from EN?

cmb69: Yes. These should be removed from translations as well.

saundefined: Looked at the language repositories - both folders are deleted in all translations.
It seems possible to remove

saundefined:
Applied with 010a8aa3015f7dcd5203f7c1e49b74d589d3aeb9 and 74194ed130e3e0e822ff8a8f10f2c5391341e748.
Thanks, everyone!

PR #27, opened 2021-05-19
Remove right to left patch

Girgias:
None of our active translation are right-to-left, and this should be handled by UTF-8 nowadays


cmb69 (approved):
Yes, this looks like a relict of long passed times.

sy-records (approved):

PR #31, opened 2021-05-24
Remove iniupdate scripts

Girgias:
I think this was meant to be used to fill in information about INI settings but hasn't been used in the last decade


cmb69:
I guess this can be removed.

PR #34, opened 2021-06-03
Add section order QA script

Girgias:
This QA scripts checks if all mandatory sections are present and that they are in the correct order, if not print the file name.

The docs stills needs fixing and the script possibly improved before we can run this in CI for doc-en.

Also just to check, @thinkverse you're fine with relicensing your code under the PHP License and merge it into the project?


thinkverse:
Thanks for the ping, and yeah it's absolutely fine, go ahead. 👍 🙂

PR #37, opened 2021-07-17
Add intersection types to our docbook dtd

cmb69:
DocBook 5.2 adds support for union and intersection types.  We already
support union types, but as of PHP 8.1.0 intersection types are
available, so we now support these as well.

Back port the syntax for intersection types to our older DocBook DTD.

Example:
`<type class="intersection"><type>Traversable</type><type>Countable</type></type>`

cc @Girgias


cmb69:
We also need support for intersection types in docgen.php, but that doesn't look urgent.

Girgias (approved):
LGTM

PR #44, opened 2021-08-28
Generate `<type>int</type>` instead of `integer`

TysonAndre:
And trim trailing whitespace

Mentioned in https://github.com/php/doc-en/pull/885#discussion_r697786840

### DIFF
--- a/scripts/docgen/docgen.php
+++ b/scripts/docgen/docgen.php
@@ -1,5 +1,5 @@
 <?php
-/*  
+/*
   +----------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


salathe:
Note, this change only affects `ini.xml` (that much wasn't clear to me from the PR, commit log, nor commit diff at first glance).  

Docgen will still generate `<type>integer</type>` elsewhere due to the use of [`gettype()`](https://www.php.net/gettype); including the specific case mentioned in https://github.com/php/doc-en/pull/885#discussion_r697786840 for the extension's constants page.

PR #45, opened 2021-08-28
Generate int/float/null instead of integer/double/NULL in docgen.php

TysonAndre:
For documentation of global constants and class constants.

Address review comments in #44


Girgias:
Maybe we could use the new PHP 8.0 function ``get_debug_type()``? Although that means a special case to handle ``resource`` and require people to use PHP 8.

TysonAndre:
> Maybe we could use the new PHP 8.0 function `get_debug_type()`? Although that means a special case to handle resource and require people to use PHP 8.

I'd rather not - I'm using 7.4 right now, OSes with Long Term Support (LTS) releases would be using php 7 for a while, and 7.3 still has security support. Some of these scripts claim to be for PHP version 5 in the file headers.

scripts/docgen/docgen.php
```diff
@@ -187,6 +187,18 @@ function save_file($filename, $content) { /* {{{ */
...
 /* }}} */
 
+function get_type_of_value($value) { /* {{{ */
+    if (is_int($value)) {
+        return 'int';
+    } elseif (is_float($value)) {
+        return 'float';
+    } elseif (is_null($value)) {
+        return 'null';
+    }
```
salathe, line 197:
```suggestion
    } elseif (is_bool($value)) {
        return 'bool';
    }
```

salathe (changes requested):
This function (`get_type_of_value()`) is only used for constants (due to the lack of type information in reflection for them), so doesn't need to "work" for absolutely all types (e.g. handling the `resource (closed)` from `gettype()`).  But it could do with also handling boolean values (to return `bool` instead of the `boolean` from `gettype()`).

Personally, I'd have gone for an array mapping from `gettype()` names to desired names.  But the `is_*()` calls work too.

As for `get_debug_type()`, I'm also not keen on introducing a PHP 8.0.0 requirement just yet, even though it may be getting towards that time (e.g. PHP 7 falls into security-only fixes in November).

scripts/docgen/docgen.php
TysonAndre: Done

scripts/docgen/docgen.php
```diff
@@ -187,6 +187,20 @@ function save_file($filename, $content) { /* {{{ */
 }
 /* }}} */
 
+function get_type_of_value($value) { /* {{{ */
+    if (is_int($value)) {
+        return 'int';
```
TysonAndre, line 192: Oh, right, I'd also seen that `&integer;` was recommended over `<type>int</type>` in review comments when I submitted PRs that used the output of this function. I should do that instead

TysonAndre:
> Personally, I'd have gone for an array mapping from gettype() names to desired names. But the is_*() calls work too.

It'd break if they ever changed to match the php7.0 real types, which is why I preferred the verbose version

scripts/docgen/docgen.php
cmb69: Hmm, I think so far we prefer the entities only for "free text", but the synopsis use explicit `<type>` elements. It might be okay to change that (although likely a lot of work for translators), but at the very least we need to coordinate with gen_stub.php which [generates `<type>` elements](https://github.com/php/php-src/blob/ef5ed06d3cc615ad3ae5c7b7d601ab9e4a35111e/build/gen_stub.php#L719-L741).

@kocsismate, thoughts?

kocsismate: Yeah, I'd appreciate if docgen generated code which is compatible with the output of gen_stub.

kocsismate (approved):

scripts/docgen/docgen.php
```diff
@@ -187,6 +187,32 @@ function save_file($filename, $content) { /* {{{ */
...
 
+/** Returns an xml type tag or an entity to use in the generated xml for a PHP constant */
+function get_xml_type_tag_or_entity($value, bool $allow_entity) { /* {{{ */
+	// Entities from entities/global.ent are recommended for freeform text.
+	if (is_int($value)) {
+		return $allow_entity ? '&integer;' : '<type>int</type>';
+	} elseif (is_float($value)) {
+		return $allow_entity ? '&float;' : '<type>float</type>';
+	} elseif (is_null($value)) {
+		return $allow_entity ? '&null;' : '<type>null</type>';
```
salathe, line 198:
```suggestion
		return '<type>null</type>';
```
Ref. https://github.com/php/doc-base/blob/e3c21d3e44e43813e9c844b33747e3a68e18f596/entities/global.ent#L556-L559

salathe:
@kocsismate can you confirm (as could probably be assumed by your approval of the PR) that these changes are, as you requested previously, compatible with the output of gen_stub?

kocsismate:
@salathe Yes, these changes are themselves are the same what gen_stub generates, however there are some other differences between the output of the two scripts which should also be addressed sometime soon. Latest example: https://github.com/php/doc-en/commit/d6aee4a5004f7d532f24c06ea2ab2ac0b91b8664#diff-6193da7441a9b194146a36138e21738e68cf63c7f17cbd4bff4726d431f2093e

PR #47, opened 2021-09-12
Add never type

kamil-tekiela:


cmb69:
Thank you! I think this can be merged right away.

PR #54, opened 2022-01-11
Replace the ftp:// URL with http as it is not easily accessible

zonuexe:
* GNU ftp page can be accessed via http
 * SunSITE.unc.edu and MetaLab.unc.edu are now called "ibiblio."
   https://en.wikipedia.org/wiki/Ibiblio
 * openldap-stable only has an old version, so replace it with another page.
   https://openldap.org/software/download/OpenLDAP/openldap-stable/


mumumu (approved):
> as it is not easily accessible

LGTM, because ftp support is removed as of Chrome 88.

https://developers.google.com/web/updates/2020/12/chrome-88-deps-rems#ftp_support_removed

cmb69:
> […], because ftp support is removed as of Chrome 88.

Firefox also dropped FTP support a while ago. :(

PR #56, opened 2022-03-15
Update natural sort algo URL

Seldaek:
http://sourcefrog.net/projects/natsort/ says it moved to GH, which is also slightly easier on the eyes 👍🏻


cmb69:
Thank you!

PR #58, opened 2022-04-04
Refactor the Check Example QA script

Girgias:
This really should do some XML parsing instead of regex guess work, but I did find a bunch of errors with it.


cmb69 (approved):
> This really should do some XML parsing instead of regex guess work, […]

Right. This way we could easily solve "do not run this script on programs which are not PHP", for instance, without excluding certain files altogether.

Still, I'm fine with sticking with regexs for now.

PR #61, opened 2022-04-28
[WIP] Add RFC skeleton for interfaces

mallardduck:
This is a WIP pull request that will have cousin PRs in varying PHP repos.

### DIFF
--- a/RFC/skeletons/interfacename.xml
+++ b/RFC/skeletons/interfacename.xml
@@ -58,8 +58,8 @@
     <xi:include xpointer="xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.baseclassname')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural') and not(modifier='private')])">
      <xi:fallback/>
     </xi:include>
-   </classsynopsis>
 -->
+   </classsynopsis>
 <!-- }}} -->
  
   </section>
--- a/RFC/skeletons/interfacename.xml
+++ b/RFC/skeletons/interfacename.xml
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="utf-8"?>
+<phpdoc:classref xml:id="class.classname"
+                 xmlns:phpdoc="http://php.net/ns/phpdoc"
+                 xmlns:xlink="http://www.w3.org/1999/xlink"
+                 xmlns:xi="http://www.w3.org/2001/XInclude"
+                 xmlns="http://docbook.org/ns/docbook">
+
+ <title>The InterfaceName interface</title>
+ <titleabbrev>InterfaceName</titleabbrev>
+ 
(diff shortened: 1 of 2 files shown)


mallardduck:
It is a little ugly, but I've gotten it working. It's really not that much uglier than the existing PHD code, and the latest scheme change I pushed here allows it be done in a way that doesn't touch existing manual pages. So the notation used by things today would still render exactly the same as it does today, to "fix" those pages they would need to use this new notation (which then triggers the new render paths).

There's only one funky bug that i'm looking through - but may take me some time to come back to sorting out. Basically with the notation I've created something is adding white space in the HTML between span tags. This results in gaps between the interface names and the `,`. Like so: Screen Shot 2022-04-28 at 5 41 06 PM

cmb69:
Hmm, I'm afraid DocBook does neither define `<interfacesysnopsis>` nor `<interfacesynopsisinfo>`. Unless we want to introduce it for our own purposes and further deviate from the standard (what I wouldn't like), we should stick with [`<classsynopsis>`](https://tdg.docbook.org/tdg/4.5/classsynopsis.html) and use the predefined `class` attribute with values `class` and `interface`, respectively. The `<classsynopsisinfo>` for interfaces likely doesn't need an attribute, as that can be determined from the context.

Anyhow, thank you very much for looking into this! :)

mallardduck:
@cmb69 - that's actually in the PR as well.

cmb69:
One of the problems with newly introduced elements is that they won't even pass the validation stage (`./configure`); we would need to adjust the DTD, and that will make updating to the latest DocBook (we're years behind) even harder, since that now uses RelaxNG (if I remember correctly).

And the required changes are presumably not that hard to do. I mean, changing the skeleton (and the existing docs) is trivial. Any changes to PhD are unlikely to be trivial (that's why I called that "ugly"), but it might not be too hard either. I'll try to have a closer look at https://github.com/php/phd/pull/64.

mallardduck:
> won't even pass the validation stage (./configure); we would need to adjust the DTD

Yeah that's what I meant - I've already updated the DTD. Ensured that the configure can run properly and then worked on making the PHD commit and PR.

The part about updating the latest DocBook I admit I'm fuzzy on, do I need to propagate any changes I made to the DTD somewhere else? I can take a stab at that if directed the right way.

In any case maybe upon further review of the PR to PHD things will seem less ugly and more acceptable. Or hopefully at the very least it will inspire you or someone more experienced on how to accomplish this.

---

Legitimately, for years I thought DS provided classes due to the docs showing them as abstract classes. So I think there's immense value here to correct and prevent future confusion. I'm a product of the Internet when half the answers were "RTFM" still, and one thing that taught me is that telling someone to RTFM only goes as far as the FM is accurate. I'm glad things have gotten a bit more welcoming than those days tho, but still a context/POV I appreciate having.

PR #66, opened 2022-07-05
Remove url.crypt.blowfish entity

kamil-tekiela:
See https://github.com/php/doc-en/pull/1667 and https://github.com/php/web-php/pull/595


cmb69:
As the CI failures show, several translations are not yet up to date for this. I suggest to wait some time before merging.

cmb69:
Seems nothing has changed till now. Maybe provide PRs for the translations? Or just break these builds?

kamil-tekiela:
> Or just break these builds?

I would be ok with that. As far as I know, it doesn't break the site, it just breaks new builds, correct?

cmb69 (approved):
> As far as I know, it doesn't break the site, it just breaks new builds, correct?

Right. And that should trigger a notification, so I think we should go ahead.

PR #70, opened 2022-10-03
Add missing entity

athos-ribeiro:
This entity was removed on 18f17b8 due to being wrongly marked as obsolete in b2bc755. The reason for it being marked as obsolete may be due to a missing `&` character to reference the URL, introduced in https://github.com/php/doc-en/commit/f1e61c544919763580fbcaac3c618db3ccdde0dd


cmb69:
Thank you!

PR #73, opened 2022-10-24
fix return true when pass htmlentities.xml

mowangjuanzi:


scripts/include/lib-translations.inc.php
```diff
@@ -87,6 +87,11 @@ function is_translatable ($filename) {
 	if (in_array(basename($filename), $files_not_translated)) {
 		return false;
 	}
+
+	// exclude filenames like htmlentities.xml
+	if (preg_match("/^entities./", basename($filename))) {
+		return false;
+	}
```
Girgias, line 94: This doesn't make sense ``htmlentities.xml`` should be marked as translatable.

mowangjuanzi: Fixed `is_translatable('htmlentities.xml')` returning `false`. should return `true`

Girgias: Okay, so the *comment* makes no sense

scripts/include/lib-translations.inc.php
```diff
@@ -87,8 +87,13 @@ function is_translatable ($filename) {
...
 		return false;
 	}
+
+	// exclude filenames like htmlentities.xml
+	if (preg_match("/^entities\./", basename($filename))) {
+		return false;
+	}
 	
-	$files_matches = array('/internals/', '/internals2/', 'entities.');
+	$files_matches = array('/internals/', '/internals2/');
```
Girgias, line 96: At this point we don't have these folders any more, so this can be removed completely.

scripts/include/lib-translations.inc.php
```diff
@@ -87,8 +87,13 @@ function is_translatable ($filename) {
 	if (in_array(basename($filename), $files_not_translated)) {
 		return false;
 	}
+
+	// exclude filenames like htmlentities.xml
+	if (preg_match("/^entities\./", basename($filename))) {
+		return false;
+	}
```
Girgias, line 94:
```suggestion
	// Exclude entity files generated by PhD
	if (preg_match("/^entities\./", basename($filename))) {
		return false;
	}
```

Girgias (changes requested):

PR #75, opened 2022-10-25
Add ext/random to manual.xml.in

TimWolla:
see php/doc-en#1916

### DIFF
--- a/manual.xml.in
+++ b/manual.xml.in
@@ -302,6 +302,7 @@
    &reference.lua.book;
    &reference.luasandbox.book;
    &reference.misc.book;
+   &reference.random.book;
    &reference.seaslog.book;
    &reference.spl.book;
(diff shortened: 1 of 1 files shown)


TimWolla:
Interesting, CI didn't fail, despite the entity not existing (yet, as expected):

https://github.com/php/doc-base/actions/runs/3324738526/jobs/5496677311#step:6:37

cmb69:
> Interesting, CI didn't fail, despite the entity not existing (yet, as expected):

Yeah, that should be investigated closer.

cmb69:
Anyway, we need to remove `&reference.csprng.book;` from this file.

TimWolla:
> Anyway, we need to remove `&reference.csprng.book;` from this file.

Of course :facepalm: Thanks!

PR #80, opened 2022-11-16
Suppress bogus "XPointer evaluation failed" libxml error

cmb69:
This error is reported even when there is an `<xi:fallback>` specified, what makes that error bogus.  If no `<xi:fallback>` is specified, there is still another error message:

````
could not load file:%s, and no fallback was found
````


Girgias:
Superseded by https://github.com/php/doc-base/pull/99

PR #82, opened 2022-12-05
Use available php for GH CI

cmb69:
Hard-coding a certain php version doesn't make sense if we use ubuntu- *latest*.  Either we use the default php (as with this patch), or we commit to a certain Ubuntu version.

---

cc @mumumu, @localheinz


localheinz (commented):
👍

mumumu (approved):
Other than a few locales with xml build errors, everything seems fine, looks good to me.

cmb69:
Thank you for review!

PR #86, opened 2022-12-16
Align the output of docgen.php to gen_stub.php

kocsismate:
I only fixed a few things, but then realized that it's not possible to completely align the output of the two scripts, since gen_stub.php has much more information about symbols than docgen.php has (like the `@readonly` or `@type` PHPDoc).


scripts/docgen/docgen.php
```diff
@@ -606,7 +609,7 @@ function gen_class_markup(ReflectionClass $class, $content) { /* {{{ */
 			}
 			$markup .= '<classsynopsisinfo role="comment">&InheritedProperties;</classsynopsisinfo>'. PHP_EOL;
 			foreach ($inherited as $declaring_class) {
-				$markup .= str_repeat(' ', $ident) ."<xi:include xpointer=\"xmlns(db=http://docbook.org/ns/docbook) xpointer(id('" . strtolower($declaring_class) . ".synopsis')/descendant::db:fieldsynopsis)\"><xi:fallback/></xi:include>". PHP_EOL;
+				$markup .= str_repeat(' ', $ident) ."<xi:include xpointer=\"xmlns(db=http://docbook.org/ns/docbook) xpointer(id('" . strtolower($declaring_class) . ".synopsis')/descendant::db:fieldsynopsis)\">" . PHP_EOL . str_repeat(' ', $ident + 1) . "<xi:fallback/>" . PHP_EOL . str_repeat(' ', $ident) . "</xi:include>". PHP_EOL;
```
Girgias, line 612: Wouldn't it be better to use ``"\n"`` instead of PHP_EOL? As the output on Windows will be different from Nix

Girgias (commented):
I never used docgen, so I prefer to see @cmb69's comments.
But it looks sensible to me.

scripts/docgen/docgen.php
kocsismate: Yeah, I was thinking about changing this, but at last I decided not to touch so many instances of `PHP_EOL` (150+). However, this could de addressed in a separate PR if Christoph also agrees.

Girgias: Right makes sense to leave it as is then

Girgias (approved):

PR #87, opened 2023-01-12
Update url.ip2long.tech

jbafford:
The old url is gone. The new url has equivalent content, though from a newer version of the source material.

Wayback machine link to old content: https://web.archive.org/web/20220108175123/http://publibn.boulder.ibm.com/doc_link/en_US/a_doc_lib/libs/commtrf2/inet_addr.htm

Fixes https://github.com/php/doc-en/issues/2168


Crell:
Thanks!

PR #91, opened 2023-02-20
Update config for `xmllint` at `xml-check.php`

phansys:
The XML files in the docs are using 1 space for indentation, so setting the `XMLLINT_INDENT= ' '` environment variable and the `--format` option for `xmllint` ensuring the right indentation makes sense to me.
Regarding the encoding, [some files](https://github.com/php/doc-base/blob/140b2f686197c7100f5126a7e2ac16fb096ab309/funcindex.xml#L1) are using `UTF-8`, [others](https://github.com/php/doc-en/blob/cb010c77b6a2f0b35218cd1663cb8f5b88035069/reference/ibm_db2/configure.xml#L1) `utf-8`.

See: https://gnome.pages.gitlab.gnome.org/libxml2/xmllint.html

### DIFF
--- a/scripts/xml-check.php
+++ b/scripts/xml-check.php
@@ -43,7 +43,7 @@ $fp = fopen($example_filename, "wb");
 fwrite($fp, $file);
 fclose($fp);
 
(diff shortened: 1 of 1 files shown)


Girgias:
Can you please explain the motivation for this change?

phansys:
>Can you please explain the motivation for this change?

Sure, I updated the PR description.

PR #95, opened 2023-08-08
Fix GH-2655: add `PEM` and `DER` acronyms

alexandre-daubois:
Related to https://github.com/php/doc-en/issues/2655

### DIFF
--- a/entities/acronyms.xml
+++ b/entities/acronyms.xml
@@ -67,6 +67,12 @@
    </listitem>
   </varlistentry>
   <varlistentry>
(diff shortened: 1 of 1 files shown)


Girgias (commented):
Does it make sense to also add PKCS (looked this up via: https://serverfault.com/a/9717)

alexandre-daubois:
Good idea! Created https://github.com/php/doc-base/pull/96, will do the PR in doc-en :slightly_smiling_face:

PR #85, opened 2022-12-13
Let build fail for erroneous xincludes

cmb69:
So far, these errors have been reported, but didn't cause the build to fail, so were easily overlooked[1].

[1] <https://github.com/php/doc-en/commit/89eaa41302bd5a9c7f66640446eefde6061649a4#r92913331>

PR #98, opened 2023-08-14
Simplify XInclude/XPointer error reporting.

alfsb:
This makes XInclude/XPointer error reporting less annoying, but doesn't omit them totally.

Impacts https://github.com/php/doc-base/pull/80 and https://github.com/php/doc-base/pull/85, end avoids the massive https://github.com/php/doc-en/pull/2015.

For https://github.com/php/doc-base/pull/85, I propose making the build fail only on `en/`, as making this a error on translations would kill (almost) them all.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -67,8 +67,10 @@ Package-specific:
   --with-php=PATH                Path to php CLI executable [detect]
   --with-lang=LANG               Language to build [{$acd['LANG']}]
   --with-partial=my-xml-id       Root ID to build (e.g. <book xml:id="MY-ID">) [{$acd['PARTIAL']}]
-  --disable-broken-file-listing  Do not ignore translated files in 
+  --disable-broken-file-listing  Do not ignore translated files in
                                  broken-files.txt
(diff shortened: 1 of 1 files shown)


Girgias (approved):
This looks sensible to me

alfsb:
By coincidence, the manual is warning a XPointer failure on `en/`
```
XPointer evaluation failed: #xmlns(db=http://docbook.org/ns/docbook) xpointer(id('class.mongodb-bson-persistable')/db:refentry/db:refsect1[@role='description']/descendant::db:methodsynopsis[not(@role='procedural')])
```
I can merge a new commit, something like:
```php
if ( count( $errors ) && $ac['LANG'] == 'en' )
    errors_are_bad(1);
```
This will make the `en/` building to fail completely in case of failed XPointer evaluations, and will permit remove all xi:failbacks for the sake of translations, as we now have `--disable-xpointer-reporting`.

May I proceed directly, or open a specific issue about this?

Girgias:
Probably open an issue first, so that this can be fixed and then make the build fail?

Girgias:
Ah it already got fixed, so yes can be enabled now

PR #108, opened 2023-12-03
Delete files with no licenses

Girgias:
@derickr are any of these files used on the build machine?


derickr:
At least one is, I can't guarantee the others, so I would be reluctant to just remove them.

```
$ grep -r fetch-ch
rsync.php.net/cronjobs.docbuilder:10 9 * * 5 php /home/docbuilder/phpdoc-all/doc-base/scripts/fetch-chms.php
docs/README.rsync:CHMS are copied from oti1.php.net with the phpdoc/doc-base/trunk/scripts/fetch-chms.php script
```

And you can't delete this one: [scripts/gen-phpdoc-tz-list.php](https://github.com/php/doc-base/pull/108/files#diff-c81ed0a05c88de0574280ec92bcb1ae523f7f9518156285c2e3287beb6291b6d) — it is used when the tz-data is updated. I wrote it, so pick what licence makes sense for it: https://github.com/derickr/timelib/blob/master/zones/Makefile#L69

PR #111, opened 2023-12-14
Revcheck library and QA tools for translation synchronization 

alfsb:
This PR rewrites revcheck logic into more modern PHP usage, and publishes a series of tools to check if one translation is in sync with the base language.

The `scripts/translation` and `scripts/translation/lib/` directories are written as self contained code ,that can be checked out or utilized as a git submodule to obtain revcheck data in typed form


scripts/translation/lib/RevcheckData.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public RevcheckFileList $sourceFiles;
+    public RevcheckFileList $targetFiles;
+
+    function __construct( $sourceDir , $targetDir )
+    {
+        $this->sourceFiles = new RevcheckFileList( $sourceDir );
+        $this->targetFiles = new RevcheckFileList( $targetDir );
+    }
+}
```
Girgias: Nit EOL before EOF

scripts/translation/lib/RevcheckFileInfo.php
```diff
@@ -0,0 +1,35 @@
...
+    function __construct( string $path , string $name , int $size , $date = new DateTime('@0') )
+    {
+        $this->path = $path;
+        $this->name = $name;
+        $this->size = $size;
+        $this->date = $date;
+        $this->key = ltrim( $path . '/' . $name , '/' );
+        $this->syncStatus = RevcheckStatus::Untranslated;
+    }
+}
```
Girgias: Ditto

scripts/translation/lib/RevcheckFileList.php
```diff
@@ -4,26 +4,30 @@
 
 class RevcheckFileList
 {
-    private $list;
+    var $list = array();
```
Girgias, line 24: Why use ``var`` instead of ``public``?

scripts/translation/lib/RevtagParser.php
```diff
@@ -57,13 +54,16 @@ public static function parseComment( DOMNode $node , RevtagInfo & $ret , $filena
 
         if ( str_starts_with( $text , "EN-" ) )
         {
+            // /EN-Revision:\s*(\S+)\s*Maintainer:\s*(\S+)\s*Status:\s*(\S+)/ // restric maintainer with no spaces
```
Girgias: "restrict" ?

scripts/translation/lib/GitLogParser.php
```diff
@@ -42,22 +42,37 @@ static function parseInto( string $lang , RevcheckFileList & $list )
             // commit message
             if ( str_starts_with( $line , '    ' ) )
             {
-                // commits marked with `[skip-revcheck]` are skiped in full
+                // commits with this mark are ignored
                 if ( stristr( $line, '[skip-revcheck]' ) !== false )
                     $skip = true;
                 continue;
```
Girgias: Maybe we should make this more lenient and accept ``[skip revcheck]``, ``[revcheck skip]``, and ``[revcheck-skip]`` ?

scripts/translation/README.md
```diff
@@ -62,15 +81,20 @@ php doc-base/scripts/translation/qaxml.t.php function
...
 php doc-base/scripts/translation/qaxml.t.php type
-
 php doc-base/scripts/translation/qaxml.t.php classsynopsis
 php doc-base/scripts/translation/qaxml.t.php constructorsynopsis
 php doc-base/scripts/translation/qaxml.t.php destructorsynopsis
 php doc-base/scripts/translation/qaxml.t.php fieldsynopsis
 php doc-base/scripts/translation/qaxml.t.php funcsynopsis
 php doc-base/scripts/translation/qaxml.t.php methodsynopsis
+
+php doc-base/scripts/translation/qaxml.t.php code
```
Girgias, line 91: Note: the code, literal, filename and varname are  in both the expected no translation and maybe translation categories.

scripts/translation/qaxml.a.php
```diff
@@ -30,33 +30,35 @@
     if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
         continue;
 
-    $intersect = array_intersect( $s, $t );
-    $onlySource = array_diff( $s , $intersect );
-    $onlyTarget = array_diff( $t , $intersect );
+    $header = true;
+    $match = array();
```
Girgias, line 47: You seem to be mixing new and "old" style syntax for creating an array, any reason?

Girgias (commented):
I had a quick glance commit by commit, just a couple of question where some might have been fixed along the way

scripts/translation/configure.php
```diff
@@ -0,0 +1,29 @@
...
+require_once __DIR__ . '/lib/all.php';
+
+if ( count( $argv ) < 2 )
+{
+    fwrite( STDERR , "  Missing paramater. Usage:\n" );
+    fwrite( STDERR , "    {$argv[0]} [lang_dir]:\n" );
+    return;
+}
+
+new RevcheckRun( 'en' , $argv[1] , true );
```
Girgias, line 29: Maybe allow a way to provide a custom location to ``en`` ?

scripts/translation/qaxml.e.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
Girgias, line 1: This one is kinda annoying for doc-fr as we have results like:

```

qaxml.e: ./fr/appendices/filters.xml

* &string; -2 +0
* &array; -7 +0
* &boolean; -2 +0
* &integer; -3 +0

  php ./base/scripts/translation/qaxml.e.php --add-ignore=./fr/appendices/filters.xml:5663bc86:3049c997

qaxml.e: ./fr/appendices/ini.core.xml

* &php.version.added; -2 +0

  php ./base/scripts/translation/qaxml.e.php --add-ignore=./fr/appendices/ini.core.xml:55daa03d:917d90e2

qaxml.e: ./fr/appendices/ini.list.xml

* &php.version.added; -1 +0

  php ./base/scripts/translation/qaxml.e.php --add-ignore=./fr/appendices/ini.list.xml:9b82c803:b84704f8

qaxml.e: ./fr/appendices/ini.sections.xml

* &Name; -0 +1
* &Changeable; -0 +1
* &Changelog; -0 +1
```

Which are not really problematic

scripts/translation/lib/RevcheckData.php
alfsb: This file doesn't exist anymore.

scripts/translation/lib/RevcheckFileInfo.php
alfsb: EOL before EOF already fixed in a posterior commit.

scripts/translation/lib/RevtagParser.php
```diff
@@ -0,0 +1,105 @@
...
+    public static function parseNodeRecurse( DOMNode $node , RevtagInfo & $ret , $filename = "" )
+    {
+        if ( $node->nodeType == XML_COMMENT_NODE )
+            RevtagParser::parseComment( $node , $ret , $filename );
+
+        foreach( $node->childNodes as $child )
+            RevtagParser::parseNodeRecurse( $child , $ret , $filename );
+    }
+
+    public static function parseComment( DOMNode $node , RevtagInfo & $ret , $filename = "" )
```
kamil-tekiela, line 64: I have two questions and I will use this line as an example.

1. Why so many of the new functions do not specify the return type? 
2. Why are you using pass-by-ref? At least in this example the pass-by-ref seems completely unnecessary. I don't think you ever need to use pass-by-ref, you can just return what you need to return.

scripts/translation/lib/RevtagParser.php
```diff
@@ -0,0 +1,105 @@
...
+
+    public static function parseFile( string $filename ): RevtagInfo|null
+    {
+        $doc = XmlUtil::loadFile( $filename );
+        $ret = new RevtagInfo;
+        RevtagParser::parseNodeRecurse( $doc , $ret , $filename );
+        return $ret;
+    }
+
+    public static function parseText( string $contents ): RevtagInfo|null
```
kamil-tekiela, line 47: How can this function return `null`?

kamil-tekiela:
Did you execute Psalm or PHPStan to make sure there aren't any obvious issues?

scripts/translation/lib/RevcheckFileList.php
alfsb: No typed arrays in PHP, and access trough `get()` to document the possible values.

Algo, the `$list` is constructed from file system, and is not expected of itens to be added or removed trom this list after the inicial loading.

scripts/translation/lib/RevtagParser.php
alfsb: Yes, fixed. As this is clearing for merging, there is translations with user names with spaces?

These comments are effectually a TODO, before implementation.

scripts/translation/lib/RevcheckFileList.php
kamil-tekiela: Why was it made public at all? You can still add the `array` type and then add proper type in docblock. Somethign like `string[]`. 

But I agree we should not use the old alias of `public` anymore.

scripts/translation/lib/GitLogParser.php
alfsb: This is easy to implement. But will further diverge this new code from the two+ other revchecking code, as it may cause the revcheck.html produced from this code diverge from other revcheck variations, making it a little harder to future migrations.

I suggest first replacing/removing duplicated code before changing it further.

Algo, the only documented way to skip is `[skip-revcheck]`.

scripts/translation/README.md
alfsb: Fixed. These categories may fluctuate between translations, or even in different partis inside a translation.

scripts/translation/qaxml.a.php
alfsb: No reason. It may reflect the other code style that I worked that day, before writing this code.

scripts/translation/configure.php
alfsb: It's easy to change this configure to accept one or two path locations, but I only tested running all other code from the parent directory of both `en` and `$LANG`. Running from other locations may surface new bugs.

If the server machines don't `cd` or `pushd` into a common checkout directory, or uses other names for `en`, I may work on after the new year.

scripts/translation/qaxml.e.php
alfsb: > Not all translations are identical, or use the same conventions. So not all scripts will be of use for all translations. The assumptions of each script are described in each file.

I don't expect this QAs scripts will be forced in all translations. I even don't expect two editors to use the very same conventions inside one manual language. And so, one editor can --add-ignore changes they don't bother or are in files of other maintainers.

These scripts exists to help translations to be more in sync with en, and less prone to breakage. They are **NOT** QA acceptance scripts.

scripts/translation/lib/RevtagParser.php
alfsb: 1. No specific reason. Feel free to patch every and all functions with their expected return types.
2. This is possibly a vestige from my local personal scripts, when they used primitives values as "return accumulators". With class this is not necessary. I may test this to remove the pass by reference syntax.

scripts/translation/lib/RevtagParser.php
alfsb: In general, there is no grantee that a particular dictionary will contain all the possible keys.

In specific, file paths are used as keys, and is very common for translations to don't have all files translated. Then, there will be keys on `en` list that does not exists in $LANG list;

alfsb:
> Did you execute Psalm or PHPStan to make sure there aren't any obvious issues?

No. I don't have any of those tools. These tools are not installed in automatic checking? If not, why not?

scripts/translation/lib/RevcheckFileList.php
alfsb: Replaced var.

scripts/translation/lib/RevtagParser.php
Girgias: I'm not sure if there are, I can run this on all of the translations to check.

scripts/translation/lib/GitLogParser.php
Girgias: Ack, I was wondering why the revcheck code was being duplicated :)

scripts/translation/configure.php
Girgias: Ack, move as a future TODO

scripts/translation/lib/GitLogParser.php
alfsb: In a distant future, this new folder can be submodules imported from doc-base and web-doc, and revcheck front-ends may use this unified revcheck back-end as the single source of revcheck data.

Sobak (commented):
I will try to look into the code more one day but I've been postponing this review long enough already, so let me start from the user perspective, i.e. how useful this is.

I ran all of the scripts in the order suggested in `scripts/translation/README.md` on the neglected Polish translation and... it has been a godsent for ensuring the quality. I've caught so many mistakes, sections that were not translated (but copy-pasted in English), sections that were omitted from the original diff generated from revcheck, translated parameters and other kinds of issues that I was surprised. After all it's a pleasant surprise because it gives more confidence for the quality of translation.

---

Speaking from my personal standpoint I would be in favor of merging it and iterating in-tree where necessary. There's one actual code complaint, though. Where ran on PHP 8.3+ there are complaints about dynamic properties in classes - I think we can just define them and call it a day since adding an attribute is probably not really justified.

Oh, and I would convert it to PSR-2/12 but it's mostly due to the fact that it's been ages since I've seen bigger codebase formatted in different style, current parenthesis spacing reminds me of WordPress :face_with_head_bandage: 

That being said, there's no code standard that would necessarily be followed for php.net's tooling so it's just a mumbling, not a real concern that affects a functionality of this PR :)

tl;dr it works and yields very useful results - I would fix deprecations, merge it, and iterate in-tree in case we want more. I don't see any problems that would outweight the benefits.

scripts/translation/lib/GitLogParser.php
alfsb: As this PR merged, I will first start a new PR to address specifically this item, and then after this, start the move to remove revcheck code duplication from doc-base.

Girgias: Thank you!

PR #115, opened 2024-01-27
Replace PHP_INI_* with INI_* constants

haszi:
Replaces the `PHP_INI_*` mode constants with their userland exposed `INI_*` equivalents as discussed [here](https://github.com/php/doc-en/pull/3140#issuecomment-1913303428).


scripts/mk_ini_set_table.sh
```diff
@@ -28,23 +28,23 @@ awk 'BEGIN {
 	print "   </thead>";
 	print "   <tbody>";
 	print "    <row>";
-	print "     <entry>PHP_INI_USER</entry>";
+	print "     <entry>INI_USER</entry>";
```
Girgias, line 31: Add constant tags?

Girgias (approved):
LGTM, thank you!

PR #116, opened 2024-01-31
Fix git refs not showing on Windows

haszi:
Fixes the issue with git ref printing on Windows showing the following error messages on each run:

```
The system cannot find the path specified.
The system cannot find the path specified.
The filename, directory name, or volume label syntax is incorrect.
doc-base:
The system cannot find the path specified.
The system cannot find the path specified.
The filename, directory name, or volume label syntax is incorrect.
en:
```

Additionally, this PR adds a switch to enable Windows to switch (ie. `cd`) between drives and not just directories on the same drive.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -27,7 +27,7 @@ echo "configure.php: $cvs_id\n";
 function usage() // {{{
 {
     global $acd;
-    
+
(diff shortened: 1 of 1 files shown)


Girgias (approved):
LGTM

PR #121, opened 2024-04-22
Add file modification history update script

haszi:
Add file modification history update script. This script can be used to update the `fileModHistory.php` files in the documentation repos which contain the last modification date/time and the contributor list for every file in the repo.

The ideal way to use this script would be something like a pre-commit hook which would generate the update to the history file, and be merged and potentially reverted together with its parent commit. If anyone knows how to do this on GitHub, please let me know.

### DIFF
--- a/scripts/updateModHistory.php
+++ b/scripts/updateModHistory.php
@@ -2,35 +2,55 @@
 
 $modHistoryFile = 'fileModHistory.php';
 
+$runningInGithubActions = (getenv("GITHUB_ACTIONS") !== false);
+
+$head = $runningInGithubActions ? "\$GITHUB_SHA" : "HEAD";
+
 $modHistoryArray = [];
 if (file_exists($modHistoryFile)) {
+    echo timeStamp() . " - Loading modification history file... ";
     $modHistoryArray = include $modHistoryFile;
     if (!is_array($modHistoryArray)) {
-        exit("Corrupted modificiation history file\n");
+		echo "file is corrupted (not an array)\n";
+        exit(1);
     }
+    echo "done\n";
+} else {
+    echo timeStamp() . " - Modification history file doesn't exist\n";
 }
 
 if (isset($modHistoryArray["last commit hash"]) && $modHistoryArray["last commit hash"] !== "") {
-    $cmd = "git rev-parse --quiet --verify " . $modHistoryArray["last commit hash"];
-    if (exec($cmd, $verifiedHash) === false) {
-        exit("Could not retrieve hash of the last commit\n");
+    echo timeStamp() . " - Found last commit hash: " . $modHistoryArray["last commit hash"] . "\n";
+    echo timeStamp() . " - Retrieving hash of the common ancestor of HEAD and the last commit... ";
+    $cmd = "git merge-base " . $modHistoryArray["last commit hash"] . " $head";
+    if (exec($cmd, $commonAncestor) === false) {
+		echo "failed\n";
+        exit(1);
     }
-    if (implode("", $verifiedHash) !== $modHistoryArray["last commit hash"]) {
-        // we cannot handle reverted commits as we don't know what changes to roll back
-        exit("Modification history file's commit hash is not in this branch's commit history\n");
-    }
-    $lastCommitHash = $modHistoryArray["last commit hash"];
+    $commonAncestorHash = implode("", $commonAncestor);
+    echo "done: ";
 } else {
+    echo timeStamp() . " - Last commit hash not found. Using empty git tree hash: ";
     // since there is no modification history, generate it for all commits since the inital one
     // 4b825dc642cb6eb9a060e54bf8d69288fbee4904 is the SHA1 of the empty git tree
-    $lastCommitHash = "4b825dc642cb6eb9a060e54bf8d69288fbee4904";
+    $commonAncestorHash = "4b825dc642cb6eb9a060e54bf8d69288fbee4904";
 }
+echo $commonAncestorHash . "\n";
+
+echo timeStamp() . " - Retrieving number of files with a diff... ";
+$cmd = "git diff --name-only $commonAncestorHash $head | wc -l";
+if (exec($cmd, $numOfFilesWithDiff) === false) {
+    echo "failed\n";
+    exit(1);
+}
+$numOfFilesWithDiff = implode("", $numOfFilesWithDiff);
+echo "done (" . $numOfFilesWithDiff . ")\n";
 
 $modifiedFilescommand = <<<COMMAND
 #!/usr/bin/env bash
 echo "last commit hash:"
-echo "$(git rev-parse HEAD)"
-git diff --name-only HEAD $lastCommitHash | while read -r filename; do
+echo "$(git rev-parse $head)"
+git diff --name-only $commonAncestorHash $head | while read -r filename; do
   echo "filename:"
   echo "\$filename"
   echo "modified:"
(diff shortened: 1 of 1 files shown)


Girgias:
@shivammathur would you possibly know how to do the git commit hooks?

shivammathur:
@Girgias @haszi 

To create a pre-commit hook, we have to create a executable `.git/hooks/pre-commit` file that runs `scripts/updateModHistory.php`.

But to enforce that this file is there on all clones, we have a couple of options.

If we can add composer to this project, we can run a post install script that copies the pre-commit file to git/hooks/pre-commit.
 
Another alternative is using https://pre-commit.com/

haszi:
@shivammathur Thanks for the explanation. 

Maybe a pre-commit hook wasn't a good idea for this problem after all. I think there is a way to achieve the same thing in the language repos by the workflow below. I do have some questions though:

Would this workflow trigger for every merge/revert on the repo? 
Would the `git commit --amend` work without any issues?

```yaml
on:
  push:
    branches:
      - master

jobs:
  if_merged:
    if: github.event.pull_request.merged == true
      runs-on: ubuntu-latest
      steps:
        - name: Checkout documentation
          uses: actions/checkout@v4

        - name: Update fileModHistory.php
          run: |
            # generate the updates to fileModHistory.php
            php updateModHistory.php

            # add and amend generated changes to the last commit
            git add fileModHistory.php            
            git commit --amend
```

shivammathur:
@haszi 

It will be better to run it on the pull_request closed event, and then check to make sure it was merged when it was closed.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

I would suggest making a new commit and pushing it to master from CI instead of amend the PR commit. Also to push it to master it will need write permission on the `GITHUB_TOKEN` secret, so configure that using the `permissions` key.

```yml
on:
  pull_request:
    types:
      - closed
    branches:
      - master
permissions:
  contents: write
jobs:
  if_merged:
    if: github.event.pull_request.merged == true
      runs-on: ubuntu-latest
      steps:
        - name: Checkout documentation
          uses: actions/checkout@v4

        - name: Update fileModHistory.php
          env:
            GITHUB_TOKEN
          run: |
            # generate the updates to fileModHistory.php
            php updateModHistory.php
            
            # Configure git
            git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
            git config --local user.name "github-actions[bot]"

            # add and amend generated changes to the last commit
            git add fileModHistory.php            
            git commit -m "Update fileModHistory.php"
        - name: Push changes
          uses: ad-m/github-push-action@master
          with:
            github_token: ${{ secrets.GITHUB_TOKEN }}
            branch: master
```

haszi:
I've discussed the adding of new commits instead of amending commits with Girgias and the `fileModHistory.php` file would probably have to be in a dedicated repo so as not to spam the documentation repos' commit history. In that scenario the workflow would still be triggered by the documentation repos after PRs are merged, which would have to pull their respective `fileModHistory.php` file from its repo, run the update script and commit the changes back to the file's repo.

I do have one question regarding the action/workflow:  would this also run the update script when commits are reverted? It would be great if the reverts could use the same workflow even though there probably aren't too many PRs reverted in the `doc-*` repos.

shivammathur:
@haszi Yes, the workflow with `pull_request` event would run on the revert PR as well.

haszi:
@shivammathur I've tested the workflow and it works like a charm. I had to add

```yaml
      with:
        fetch-depth: 0
```

under the "Checkout documentation" step to make my script work (it needs the entire git history) and the file modification info file is now being generated perfectly in the same repo.

Could you also point me in the right direction on how to check out the `fileModHistory.php` file from another repo, run the script to make updates and push the file back to the repo it was checked out of?

haszi:
@Girgias I opened this PR up for review as the script itself is ready. The changes in the last commit make the script accept both the path of the documentation and that of the "history file" as command line parameters, so this can be called (directly or from a GitHub workflow) from within the same or another repo.

scripts/updateModHistory.php
```diff
@@ -0,0 +1,219 @@
...
+        case "modified:":
+            $currentType = "modDateTime";
+            return;
+        case "contributors:":
+            $currentType = "contributors";
+            return;
+        case "last commit hash:":
+            $currentType = "commitHash";
+            return;
+    }
```
Girgias, line 159: Maybe use `match` here? Or is it expected to not be one of these cases?

scripts/updateModHistory.php
```diff
@@ -0,0 +1,219 @@
...
+function verifyCommandLineOptions($commandLineOptions): void {
+    echo timeStamp() . " - Parsing command line arguments... ";
+    if ($commandLineOptions === false
+        || !isset($commandLineOptions["docs-path"])) {
+        echo "\"--docs-path\" is a required argument\n";
+        exit(1);
+    }
+    echo "documentation path supplied\n";
+    if (isset($commandLineOptions["history-path"])) {
+        echo "                                               mod history file path supplied\n";
```
Girgias, line 195: Maybe have a variable for the space indent?

Girgias (commented):
I haven't tried it locally yet, but some minor nits.

Also @cmb69 @jimwins may want to have a look

scripts/updateModHistory.php
haszi: It should only be one of these cases and everything else is skipped by the function.

cmb69 (commented):
Sorry, I do not understand the purpose of having a fileModHistory.php file.

cmb69:
Thanks @haszi for the explanation! While this is low priority for me, I understand that it will be more important for others, so I'm generally fine with the approach.

scripts/updateModHistory.php
haszi: I just realized that I was wrong: this should be a `switch` so the function can return as `$line` won't be any of the above cases most of the time but will contain all modification dates/times and the names of every commit author.  Ie. when `$currentType` is set but `$line` is not any of those cases, it contains the information we are actually looking for (filenames, dates/times, commit authors).

scripts/updateModHistory.php
haszi: I used `strlen()` in my last commit. Will that work too or would a variable be better in this case?

scripts/updateModHistory.php
```diff
@@ -0,0 +1,219 @@
...
+        echo "documentation path \"" . $commandLineOptions["docs-path"] . "\" doesn't exist\n";
+        exit(1);
+    } else if (!is_dir($commandLineOptions["docs-path"])) {
+        echo "documentation path \"" . $commandLineOptions["docs-path"] . "\" is not a directory\n";
+        exit(1);
+    }
+    echo "documentation path verified\n";
+
+    if (isset($commandLineOptions["history-path"])) {
+        echo "\n" . str_repeat(" ", strlen($output)) . "mod history file path ";
```
cmb69, line 209: Possible alternative:
```suggestion
        printf("\n%*smod history file path ", strlen($output), " ");
```

haszi: Done

scripts/updateModHistory.php
```diff
@@ -0,0 +1,219 @@
...
+$fp = fopen($modHistoryFile, "w");
+fwrite($fp, "<?php\n\n/* This is a generated file */\n\nreturn [\n");
+foreach ($mergedModHistory as $fileName => $fileProps) {
+    if ($fileName === "last commit hash") {
+        fwrite($fp, "    \"last commit hash\" => \"" . implode("", $fileProps) . "\",\n");
+        continue;
+    }
+    $newModHistoryString = '    "' . $fileName . "\" => [\n";
+    $newModHistoryString .= "        \"modified\" => \"" . ($fileProps["modified"] ?? "") . "\",\n";
+    $newModHistoryString .= "        \"contributors\" => [\n";
```
Girgias, line 118: Use sprintf or var_export here?

scripts/updateModHistory.php
```diff
@@ -0,0 +1,219 @@
...
+    if (isset($fileProps["contributors"])) {
+        if (!is_array($fileProps["contributors"])) {
+            exit("Non-array contributors list\n");
+        }
+        foreach ($fileProps["contributors"] as $contributor) {
+            $newModHistoryString .= "            \"" . $contributor . "\",\n";
+        }
+    }
+    $newModHistoryString .= "        ],\n";
+    $newModHistoryString .= "    ],\n";
```
Girgias, line 128: Ditto

cmb69 (commented):
I guess this is good; can't test anyway, since I cannot afford a bash. :)

scripts/updateModHistory.php
haszi: Something like this?

```php
$newModHistoryString = sprintf("%4s\"%s\" => [\n%8s\"modified\" => \"%s\",\n%8s\"contributors\" => [\n", 
    " ", 
    $fileName, 
    " ", 
    $fileProps["modified"] ?? "",
    " "
);
```

scripts/updateModHistory.php
haszi: Will this work?

```php
$newModHistoryString .= sprintf("%8s],\n%4s],\n", " ", " ");
```

scripts/updateModHistory.php
Girgias: Possibly, but this really minor actually.

PR #128, opened 2024-06-14
Remove phpdoc DTD support

Girgias:


alfsb:
This is not already completed?

PR #136, opened 2024-07-23
Remove XForms from the manual (part of php/doc-en#3579)

jimwins:
The XForms content will also need to be removed from each of the `doc-$LANG` repos.

### DIFF
--- a/manual.xml.in
+++ b/manual.xml.in
@@ -112,7 +112,6 @@
   &features.http-auth;
   &features.cookies;
   &features.sessions;
-  &features.xforms;
   &features.file-upload;
   &features.remote-files;
   &features.connection-handling;


jimwins:
Thinking about it further, the correct order to do everything would be to remove all of the references to XForms from all of the translations (references in translated `chapters/tutorial.xml` and all translated `features/xforms.xml`) and the reference to `features/xforms` from `chapters/tutorial.xml` in php/doc-en, then remove it from `manual.xml.in` here in php/doc-base, and then finally remove `features/xforms.xml` from php/doc-en. The entities can finally be removed here. That should leave everything buildable after each PR. (Unless the build fails when there are unused entities?)

Girgias:
Unused entities are fine.

You will need to do the doc-en changes first, as the translators need to update the `EN-Revision` tag, I have pushed major changes by using a revcheck skip but I try to limit my usage of those.

jimwins:
Going to go ahead and merge this now that it's just three languages (it, pl, tr) that haven't pushed the submitted PR to remove this. They can clean up after the fact.

PR #137, opened 2024-07-24
Fix broken URL to O'Reilly book

jimwins:
Just happened to stumble across this. All of the external URLs should probably be reviewed.


cmb69 (approved):
Thank you!

PR #143, opened 2024-08-01
Drop url.pecl.windows.releases entity declaration

cmb69:
If https://github.com/php/doc-en/pull/3633 will be merged, there are no remaining uses of that entity in doc-en, so after translations have caught up, the entity declaration can be removed.


philip:
There was one remaining reference in `install/windows/manual.xml` that's now gone, so it'll be awhile longer.

alfsb:
There is no need to wait for translations to catch up. Any entity removed from doc-base or doc-en can be "moved" to doc-en/entities/entities.remove.ent , or declared empty there, so translations keep building _and_ are alerted about removed entities.

There is a reported issue where entity references in entities contents are being lost, something that I try to fix in the next few days.

alfsb:
`doc-ru` still uses it. I will "move" it into entities.remove.ent, so this can be merged before convertion to XML start.

alfsb:
This can now be merged.

As note above, and demonstred in PR 5704 also linked above, now it is possible to remove DTD entities from doc-en in one go, be the entity declared in doc-en or doc-base, without breaking translations.

Happy DTD entity hacking!

PR #145, opened 2024-08-07
Handle extensions state

jimwins:
Related to php/doc-en#3643.

### DIFF
--- a/scripts/qa/extensions.xml.php
+++ b/scripts/qa/extensions.xml.php
@@ -84,9 +84,15 @@ foreach ($files as $filename) {
 	}
 
 	if (preg_match('/<!-- State: (\w+)/S', $file, $match)) {
-		$State[$match[1]][$ext] = 1;
+		switch ($match[1]) {
+		case 'experimental':
+		case 'deprecated':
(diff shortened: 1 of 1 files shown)


cmb69:
Great! Thank you @jimwins!

Note though that [checking out to `en` is customary](https://github.com/php/doc-base/blob/master/README.md#quick-reference), and looking for `doc-en` breaks the usual expectations.

I tried the script without the doc-en changes, and it works fine.

cmb69:
@Girgias, since you've worked on that script, you may want to have a look.

jimwins:
Okay, rebased without the `doc-en` change and fixed up my local setup.

Girgias (approved):
Did a quick review before I'm flying, but this looks sensible

Girgias:
I'm confused why the CI is complaining about the extensions.xml file not being up to date

cmb69:
I think the script would need to be run without the `--check` flag.

jimwins:
Because now that the script handles State, doc-en *is* out of date. Thus php/doc-en#3645.

cmb69:
And doc-en is green!

PR #146, opened 2024-08-22
Only include per-function changelog since 7.0.0, not 5.0.0

jimwins:
Related to php/phd#16, which fixes the changelog generation.


Girgias:
The thing is that the changelogs should be removed in the doc sources rather than remove the generation IMHO.

jimwins:
Turns out all of the 5.x changelogs were already removed. Submitted a PR to stamp out the last few 4.x that lingered.

https://github.com/php/doc-en/pull/3775

PR #147, opened 2024-08-23
Add entities for Composer/Packagist, and new install.composer chapter

jimwins:
Related to php/doc-en#3677.


declounsher:
`[</dix>]

PR #152, opened 2024-09-14
Remove url.readline

petk:
This URL is being removed from the docs in favor of the libedit library, a non-GPL replacement for the GNU Readline library.


cmb69:
Maybe wait with merging until at least some of the translations have been updated.

philip:
The entity is still used in `reference/readline/setup.xml`, so instead of removing the entity definition consider using a different entity in `reference/readline/setup.xml`?

Girgias:
This seems to not even work for doc-en, thus closing.

Issue #153, opened 2024-09-23
broken link


cristicotet:
From manual page: https://php.net/class.libxmlerror

---
the link to official error codes is broken

cmb69:
Thanks for having reported this issue, @cristicotet!

PR #154, opened 2024-09-23
Update xmlsoft.org URLs to gitlab.gnome.org

cmb69:
The old Website is still HTTP only, but redirects to the new Website; this may not work without issues, though.

Closes GH-153.


cmb69:
it CI failure is unrelated.

PR #157, opened 2024-09-27
Remove the Installation problems chapter

jimwins:
References should link to faq.installation instead (related to php/doc-en#3785)

### DIFF
--- a/manual.xml.in
+++ b/manual.xml.in
@@ -61,7 +61,6 @@
   &install.cloud.index;
   &install.fpm.index;
   &install.pecl;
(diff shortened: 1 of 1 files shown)


Girgias:
@saundefined seems doc-ru hasn't applied the corresponding doc-en change.

saundefined:
@Girgias weird, I can't stop workflow to re-run it (or rather I can, but it doesn't stop :) )

I updated the doc-ru and applied this patch locally,
the build was successful.

Thanks!

PR #161, opened 2024-10-05
Stop bugging Harald

cmb69:
See <https://github.com/php/doc-en/pull/3836>.

---

Likely fails for some translations. Let's see.


kamil-tekiela (approved):
Should we update all translations first?

cmb69:
> Should we update all translations first?

I wouldn't know how to do that (except for German). Maybe we should just wait some days or weeks, and merge then.

Girgias:
> > Should we update all translations first?
> 
> I wouldn't know how to do that (except for German). Maybe we should just wait some days or weeks, and merge then.

I thinking waiting for a couple of days and restarting the CI is a better idea. :)

cmb69:
Just triggered CI again; most translations fail because of the MongoDB set to book conversion. I don't think adding this on top would be an issue.

PR #163, opened 2024-10-09
Fix CHM downloads due to certificate weirdness (fixes php/systems#58)

jimwins:


jimwins:
Whoops, branched from another branch instead of master. Fixed.

cmb69:
Hmm, I'm slightly uneasy to deactivate the certificate check, and after all, the latests CHMs are only from Aug 17th.

Maybe we should wait a bit. Hopefully @mumumu is well.

mumumu:
Sorry for too late reply.

I have opened port 80 of my host, so this PR can be closed if you change the following URL to `http://www.mumumu.org/chm`

https://github.com/php/doc-base/blob/master/scripts/fetch-chms.php#L3

jimwins:
Well, looks like it's still not working with this change. Someone with access to rsync.php.net, like @derickr, will need to debug.

jimwins:
And now they are flowing again, so either I just wasn't patient enough or someone loosened the gears. Latest CHM downloads re from 18 Oct 2024.

PR #170, opened 2024-10-23
Use RelaxNG for validation instead of DTD for DocBook 5.2

Girgias:
Requires https://github.com/php/doc-en/pull/3911


jimwins (approved):
MSTM, and I can confirm that it reports errors when I try to build the `doc-en` tree as of right now using the Makefile/docker setup.

cmb69:
I think we need some proper error reporting for CI at least. Can we run Jing there if the build step failed?

cmb69:
> I can confirm that it reports errors when I try to build the `doc-en` tree as of right now using the Makefile/dock

Ah, already tracked as https://github.com/php/doc-en/issues/3912.

cmb69 (commented):
I think we also should add a license file for jing. The downloads have doc/copying.html, were we like should strip the `<h2>Third-party JARs</h2>` part.

Girgias:
I am going to need to fix the translations first, as they are violating RelaxNG.

Also we need to add Java to the CI

cmb69:
Isn't Java preinstalled on Ubuntu runners? (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md)

Girgias:
> Isn't Java preinstalled on Ubuntu runners? (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md)

Yeah, no the issue is the usage of relative paths

Girgias:
Well, need to wait for translation to sync with EN with regard to the entity removal of 33ef031b1c248778d188b0ab518e317273e3f867 failure.

cmb69:
Bigger problem is likely the MongoDB set to book conversion, although there should be PRs.

Girgias:
doc-it doesn't have the MongoDB problem, but is quite out of date and doesn't build any more even with some syncing.

Might need to go back to deleting some files for it to build....

cmb69:
Looking at the recent commit history of doc-it, it might make sense to disable it for the time being.

Girgias:
Doc-fr is now RelaxNG compliant,

Submitted some PRs to doc-ja and doc-ru but waiting for them to catch up on the DOM doc-en RelaxNG violation fixes too.

Girgias:
`doc-ru` will be RelaxNG compatible the moment `reference/dom` is synced with EN (@saundefined)

`doc-ja` will be RelaxNG compatible the moment `reference/dom` is synced with EN and https://github.com/php/doc-ja/pull/135 is merged (@mumumu)

I can probably fix doc-it by using DeepL and asking some people that I know to review it.

doc-tr, seems to be some misapplied changes and requires markup fixes, which I can probably do.

`doc-uk` needs `install/unix/ngnix.xml` to be synced to fix RelaxNG violations, but it needs `language/oop5/visibility.xml` to be synced just to fix the current DTD build.

saundefined:
> `doc-ru` will be RelaxNG compatible the moment `reference/dom` is synced with EN (@saundefined)

Thanks, fixed

jimwins (approved):

Girgias:
I'll help doc-ja to get back to building.

alfsb:
`doc-pt_BR` is RelaxNG validating ok? I tested it now, and appears to be good, but to make sure.

Aside: the scripts on `doc-base/tree/master/scripts/translation` (that doc-pt_BR uses) are somewhat focused in maintaining an translation in structural sync with `doc-en`, so it may help.

cmb69:
> `doc-pt_BR` is RelaxNG validating ok?

Yes, looks good: https://github.com/php/doc-base/actions/runs/11615325856/job/32345629762

Girgias:
Yes the only ones which don't are `doc-es` and `doc-uk` (`doc-ja` has a PR to make it green)

kamil-tekiela:
I don't like this change because I don't have Java installed locally and without it I have zero information what's wrong. Previously, at least, I got some trace information that could help me debug it. Also, the parameter `--enable-xml-details` is now silently a no-op. Couldn't we at least have both options available? Make it an opt-in or opt-out.

Girgias:
> I don't like this change because I don't have Java installed locally and without it I have zero information what's wrong. Previously, at least, I got some trace information that could help me debug it. Also, the parameter `--enable-xml-details` is now silently a no-op. Couldn't we at least have both options available? Make it an opt-in or opt-out.

Unless you mess up the XML and the tree is not well balanced 95% of reasons why the validation would fail is due to a RelaxNG violation. And without a Jing run you will not be checking for schema validation errors.

Java is effectively a required dependency now.

Issue #175, opened 2024-11-10
Revcheck changes


alfsb:
With the imminent implantation of https://github.com/php/doc-base/pull/174 , I would like to make some changes. This issue discusses some minor issues that I think could be discussed in block.

### `IGNORE_EXTENSIONS_XML`

The actual revcheck ignores files named `extensions.xml` that are autogenerated. But the way the ignore are constructed, the code also ignores some non-autogenerated, translatable files.

- appendices/migration56/extensions.xml -- existent, translatable, not tracked;
- install/windows/legacy/extensions.xml -- deleted on en/, may exist untracked on translations.

This is documented as `IGNORE_EXTENSIONS_XML` on PR above, and changing/removing this in the code will cause the files mentioned above to be tracked.

### `STATUS_COUNT_MISMATCH`

The actual revcheck counts "outdated" files in two different ways;

- [Only TranslatedOld](https://github.com/php/doc-base/blob/747c53bf8ec72f27ac1a83ba91bcc390eea2e46a/scripts/revcheck.php#L603)
- [TranslatedOld + RevTagProblem](https://github.com/php/doc-base/blob/747c53bf8ec72f27ac1a83ba91bcc390eea2e46a/scripts/revcheck.php#L134)

RevTagProblem is counted towards as Old in one case, but files are shown in a separated listing (because it is impossible to generate diffs with invalid hashes). As aside, NotInEnTree is not counted at all in first case, and files in this situation are counted in the second case (default). 

This causes a mismatch between translators totals and file summary, and other mismatch between Old counts and file listings.

To resolve this, I propose to change the "wip" column in Translator's summary to "misc", and so any status other than "ok" and "old"
to be counted here.
- `ok` : files that are `ready` and hashes matches: count and listing for ok cases matches;
- `old`: files that are `ready` and hashes mismatch: count and listing for old cases matches;
- `misc`: any other case.

### `IGNORE_ONLY_WS_CHANGE` (not commited)

On discussion about [better diffs](https://github.com/php/web-doc/issues/57#issuecomment-2466712324), I observed that is possible to analyse files changes to determine it has changed only in terms of ws, automatically with `git diff -b`, or by processing text files. This can alleviate the translation's efforts in two ways:

1. To ignore "uninteresting" whitespaces changes, even if `[skip-revcheck]` was **not** utilized;
2. To ignore "uninteresting" whitespaces changes buried in squashed commits, even if `[skip-revcheck]` **was** utilized.

[...]

PR #176, opened 2024-11-11
Remove backported revcheck compatibility

alfsb:
This PR removes some of backported revcheck behaviors, that seems incorrect or makes count debugging difficult. They *new* behaviors are:

1. Tracks `chmonly/` as translatable;
2. Tracks `extensions.xml` as translatable;
3. Counts and lists ok and old exclusively.

The first change will include 4 files as translatable, the second will mark one file as translatable and possibly a few as NotInEn. The last does not change file tracking or listing, but it causes a lot of changes on translator listing.

What it does is counting Ok and Old cases in separated counters, and push all other situations on a "Other" counter. So after this, file count and file listing of outdated cases will match, and RevtagProblem will not be counted as Outdated.

Note that RevtagProblem cannot be listed alongside with Outdated anyways, because it's not possible to generate diff links in this case. This changes will make count debug vastly easier, as it removes all instances of *possibly not* counted cases in a lot of corner cases.

Reviews and commentary welcome. I plan to push this PR next week if there are no objections.

This PR supersedes https://github.com/php/doc-base/issues/175.


scripts/translation/lib/RevcheckIgnore.php
```diff
@@ -57,13 +57,6 @@ public static function ignore( $filename ) : bool
         if ( $filename == "reference/datetime/timezones.xml" )
             return true;
 
-        if ( IGNORE_EXTENSIONS_XML )
-            if ( str_ends_with( $filename , '/extensions.xml') ) // track/count backport
-                return true;
```
Girgias, line 62: Is it possible to hardcode this like the `reference/datetime/timezones.xml` file as:
```php
if ( $filename == "appendices/extensions.xml" )
            return true;
```
?

scripts/translation/lib/RevcheckIgnore.php
```diff
@@ -57,13 +57,6 @@ public static function ignore( $filename ) : bool
...
-        if ( IGNORE_CHMONLY_DIR )
-            if ( str_starts_with( $filename , 'chmonly/') )      // track/count backport
-                return true;
```
Girgias, line 65: Similarly can we hardcode the 4 files?

scripts/translation/lib/RevcheckIgnore.php
alfsb: I'm trying to make exceptions as few as possible. And in this case, I also think that the original ignore was as mistake:
```
$ find * | grep "/extensions.xml"
appendices/migration56/extensions.xml
appendices/extensions.xml
```
All files ending in "/extensions.xml" are translatable.

scripts/translation/lib/RevcheckIgnore.php
alfsb: In the same theme of having as few exceptions as possible, this change makes `chmonly/` a normal translatable directory, as it seems to me (albeit the name). This directory only contains the four .xml files, all translatable.

It is no different from `chapters/` or ` faq/` in that matter. And with CHM manuais building again, the files in this directory are being included, untranslated, for all translations.

scripts/translation/lib/RevcheckIgnore.php
Girgias: `appendices/extensions.xml` should NOT be translated, it is a generated file. Which is identical to `reference/datetime/timezones.xml` this is why I asked to hardcode the exception.

I agree that not marking `appendices/migration56/extensions.xml` as translatable is a bug.

scripts/translation/lib/RevcheckIgnore.php
Girgias: I don't think CHM builds other than English are provided, thus I don't see the point in making those files available for translation. @jimwins do you know if CHM builds can work with translations.

jimwins: We do build CHM files for other languages (not Italian and Ukrainian for some reason). @mumumu does the builds, maybe they'd know if translated files in `chmonly` would be picked up. It seems to me that they should be, but I can't build the CHM files so I can't confirm.

scripts/translation/lib/RevcheckIgnore.php
alfsb: Oh, I see the confusion. `appendices/extensions.xml` is already ignored (see line 53), and removing the blank ignore on `/extensions.xml` will not change this. So it's already done.

scripts/translation/lib/RevcheckIgnore.php
alfsb: I downloaded the chm file for Brazilian Portuguese from https://www.php.net/download-docs.php , extracted it and grep'ed it for `The Windows HTML Help Edition of the PHP Manual`, with a match. So this files are being included on translated chm manual builds. The page indicates builds as recent as 8 Nov 2024, if I ready ir correctly.

alfsb: Update. I now confirmed all 4 files of `chmonly` are included and linked on internal CHM spine and HTML rendered ToC.

Girgias: ACK, then this is fine.

alfsb:
Merged. Now, local revcheck will slight differ from hosted revcheck, as it is tracking `chmonly` and some `reserved.xml` files, and also counting changes on translator summary.

I do plan make two, possibly three changes (robust log parsing, strict rev-skip positioning, visual tweaks), beyond the already opened PR for sequential skip-revcheck.

Let me know if systems people want to wait to all changes before updating, or want to mentioned or have issues opened in each instance.

PR #184, opened 2024-11-20
Remove legacy install files.

alfsb:
I found these untranslatable files while doing some changes in `doc-base/configure.php`. They don't seem to be used or English or translations.


Girgias (approved):
I was wondering what they were used for.

PR #188, opened 2024-11-24
Use rfc-editor.org domain for RFCs

Girgias:


jimwins (approved):
LGTM

cmb69 (commented):
`https://datatracker.ietf.org/doc/html/rfc` might be even better.

PR #189, opened 2024-11-25
fix: Add hasAttributes check in XML attribute extraction

tal7aouy:
Add defensive check using ```hasAttributes()``` before iterating over 
element attributes in extractTriple function. This prevents 
unnecessary iteration and potential errors when processing XML 
elements without attributes.

Before:
- Directly iterates over attributes
- Could cause warnings on elements without attributes

After:
- Checks for attributes first
- Skips elements without attributes
- More efficient processing

### DIFF
--- a/scripts/translation/qaxml.a.php
+++ b/scripts/translation/qaxml.a.php
@@ -74,11 +74,24 @@ foreach ( $qalist as $qafile )
         print "\n";
 }
 
-function extractTriple( array $list )
+function extractTriple(array $list)
 {
     $ret = array();
-    foreach( $list as $elem )
-        foreach( $elem->attributes as $attrib )
-            $ret[] = "{$elem->nodeName} {$attrib->nodeName} {$attrib->nodeValue}";
+    try {
+        foreach ($list as $elem) {
+            if ($elem->hasAttributes()) {
+                foreach ($elem->attributes as $attrib) {
+                    $ret[] = sprintf(
+                        "%s %s %s",
+                        $elem->nodeName,
+                        $attrib->nodeName,
+                        $attrib->nodeValue
+                    );
+                }
+            }
+        }
+    } catch (Exception $e) {
+        throw new Exception("Error extracting attributes: " . $e->getMessage());
+    }
     return $ret;
 }
--- a/scripts/translation/qaxml.a.php
+++ b/scripts/translation/qaxml.a.php
@@ -77,21 +77,18 @@ foreach ( $qalist as $qafile )
 function extractTriple(array $list)
 {
     $ret = array();
-    try {
-        foreach ($list as $elem) {
-            if ($elem->hasAttributes()) {
-                foreach ($elem->attributes as $attrib) {
-                    $ret[] = sprintf(
-                        "%s %s %s",
-                        $elem->nodeName,
-                        $attrib->nodeName,
-                        $attrib->nodeValue
-                    );
-                }
+    foreach ($list as $elem) {
+        if ($elem->hasAttributes()) {
+            foreach ($elem->attributes as $attrib) {
+                $ret[] = sprintf(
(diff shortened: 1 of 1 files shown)


scripts/translation/qaxml.a.php
```diff
@@ -74,11 +74,24 @@
...
+                        "%s %s %s",
+                        $elem->nodeName,
+                        $attrib->nodeName,
+                        $attrib->nodeValue
+                    );
+                }
+            }
+        }
+    } catch (Exception $e) {
+        throw new Exception("Error extracting attributes: " . $e->getMessage());
```
alfsb, line 94: Catching and rethrowing an exception with only its message is wrong, as it erases/moves the call stack information of the original exception.

scripts/translation/qaxml.a.php
```diff
@@ -74,11 +74,24 @@
...
-function extractTriple( array $list )
+function extractTriple(array $list)
 {
     $ret = array();
-    foreach( $list as $elem )
-        foreach( $elem->attributes as $attrib )
-            $ret[] = "{$elem->nodeName} {$attrib->nodeName} {$attrib->nodeValue}";
+    try {
+        foreach ($list as $elem) {
+            if ($elem->hasAttributes()) {
```
alfsb, line 82: I would like to see some performance measurements here. See caveat in Conversation.

alfsb (commented):
While reviewing this PR, I found bugs on relevant code. This will need wait the merging of https://github.com/php/doc-base/pull/190 before continuing.

That said, I tried measure the performance impact and got almost no difference. Running it on `doc-fr`, a fully translated language (most exercite the code), resulted on timings of 
0m3,204s (old) and real	0m3,195s (new).

alfsb:
Also, I'm interested in the reasoning for `if ($elem->hasAttributes())`. The `extractTriple()` function is a private function of this file, that is called only with the results of `listNodeType( ... , XML_ELEMENT_NODE )`. If `extractTriple()` receives something that does not have an `attributes` property, this is an error that should not be silenced/ignored behind an `if()`.

Girgias:
Yeah I want to see some benchmarks here, because otherwise I don't really see an issue with the current code.

tal7aouy:
Thank you for the feedback! I understand your point regarding the performance impact, and I agree that the timing difference is negligible in this case.

The reason I kept the `hasAttributes()` check is to ensure that the `foreach` loop doesn't run on elements that don't have any attributes, which could otherwise lead to errors or unexpected behavior. Even though `extractTriple()` is only called with `XML_ELEMENT_NODE` elements, I wanted to make sure that we handle cases where an element might not have attributes gracefully, especially if the function were ever to be used in a broader context.

If performance becomes a concern in the future, I’m open to revisiting this and testing with more extensive datasets to see if there's any significant impact. Let me know your thoughts!

Girgias:
In that case I would just prefer a plain old assertion using `assert()` rather than this useless if.

tal7aouy:
@Girgias Thank you for the feedback! I will update the code to replace the if ($elem->hasAttributes()) check with an assert($elem->hasAttributes()) statement, as suggested. This enforces the assumption that all elements passed to extractTriple() have attributes, aligning with the expected usage of the function while simplifying the code.

alfsb:
> This enforces the assumption that all elements passed to `extractTriple()` have attributes, aligning with the expected usage of the function while simplifying the code.

This is neither the assumption or the expected usage of this function. Thanks for PR. but I'm calling this off. Overcomplicating the code for no performance improvement, or guarding and silencing errors that do not occur, does not improve the code.

(and I can relate where this comes from... I suffered enough with the non-orthogonal design of w3c's DOM to be automatically wary of any DOM/XML iterating code that does *not* contain all sorts of tests of corner cases.)

PR #196, opened 2024-12-03
Document and complement XInclude failures as warnings on translations

alfsb:
This PR better document how XInclude failures are treated on translations. It only cares for `<xi:include>` **without** `<xi:failback>`, so it leaves open `<xi:failback>` usage for other situations.

In fact, I tested this on all languages with `doc-en` stripped of `<xi:failback>`.

As an recommendation, I suggest mass removing all mentions of `<xi:failback>` on `doc-en`, down to XML comments, after 8.4 doc changes calms down, and possibly further creating an github action/trigger that recuses usage of `<xi:failback>`, linking this trigger to mention `XInclude failures are soft errors on translations`. on `configure.php`. The reason being that failed XInclude leaves different artifacts in the presence and absence of `<xi:failback>`. Caring for both is very confusing.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -816,8 +816,8 @@ if ( $ac['XPOINTER_REPORTING'] == 'yes' || $ac['LANG'] == 'en' )
 
 if ( $ac['LANG'] != 'en' )
 {
-    // XInclude failures are soft erros on translations, so replace
-    // residual XInclude tags on translations to keep then building.
+    // XInclude failures are soft errors on translations, so remove
+    // residual XInclude tags on translations to keep them building.
 
     $explain = false;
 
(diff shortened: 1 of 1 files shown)


alfsb:
I would like comments to making code comments and messages into idiomatic English (not my first language).

Also, I was thinking in placing some default message in the patched up nodes, something like `%xinclude fail%`, that is something small but visible, to indicate patched up nodes into rendered manual. The actual implementation creates empty tags, that may be more confusing that an small error message.

Girgias (approved):
MSTM.

alfsb:
Changed the texts a little, I plan to merge this tomorrow.

alfsb:
Merged. I expect no breakages, but let me know if are any problems. This enable removing empty `<xi:fallback>` on all languages, as they are now, and allows for `<xi:fallback>` be used for other cases.

PR #197, opened 2024-12-03
Drop url.gmagick.winbuilds

kamil-tekiela:


Girgias:
I'd wait a bit for translations to be able to catch up.

alfsb:
I put an small support for this on PR 183. Entities can be moved to an [`remove.ent`](https://github.com/php/doc-base/pull/183/files#diff-1611e343e8c176020295cf45ffbd9c767e9deebe6e507209f7e3873a64a5241bR118) file, that keeps generating entity (so translations do not break), and starts generating an statistic of [removed but still used](https://github.com/php/doc-base/pull/183/files#diff-1611e343e8c176020295cf45ffbd9c767e9deebe6e507209f7e3873a64a5241bR131) entities on each configure.

kamil-tekiela:
@Girgias Is this enough time passed already?

PR #206, opened 2024-12-12
Include Jing license that I forgot to add when commiting the JAR

Girgias:
See: https://github.com/php/doc-en/issues/3912#issuecomment-2523399759


cmb69 (approved):
Thank you!

PR #207, opened 2024-12-13
Windows related changes, ahead/behind only current branch

alfsb:
This PR makes `languages.php` more functional on Windows, and also changes repository status to not show ahead/behind of unrelated branches.

As a heads up, also remove a ugly hack for dirs named `PDO` vs `pdo`. If you are on Windows, and start receiving errors of undefined entities that contains `PDO` (upper case), you may need to filesystem remove all files from `en/reference/pdo*`, and then `git restore` then, as git on Windows may not honor case changing moves.

### DIFF
--- a/scripts/file-entities.php
+++ b/scripts/file-entities.php
@@ -18,7 +18,7 @@
 # Description
 
 This script creates various "file entities", that is, DTD entities that
-points to files and file listings, named and composed of:
+point to files and file listings, named and composed of:
 
 - dir.dir.file : pulls in a dir/dir/file.xml file
 - dir.dif.entities.dir : pulls in all files of dir/dir/dir/*.xml
--- a/scripts/file-entities.php
+++ b/scripts/file-entities.php
@@ -17,11 +17,11 @@
 
 # Description
 
-This script creates various "file entities", that is, antities and files
-that define DTD <!ENTITY name SYSTEM 'path'>, named and composed of:
+This script creates various "file entities", that is, DTD entities that
+points to files and file listings, named and composed of:
 
 - dir.dir.file : pulls in a dir/dir/file.xml file
-- dir.dif.entities.dir : pulls in a entity list for dir/dir/dir/*.xml
+- dir.dif.entities.dir : pulls in all files of dir/dir/dir/*.xml
 
 In the original file-entities.php.in, the files are created at:
 
@@ -44,7 +44,8 @@ output an solid XML grouped file (per github.com/php/doc-base/pull/183)
 so it would be possible to detect accidental overwriting of structural
 entities, the "list of entities" moved to/as normal entity text. PS: This
 will NOT work, with libxml recusing to load .manuxal.xml.in because of an
-"Detected an entity reference loop", that does not exist. Sigh.
+"Detected an entity reference loop", that does not exist. Sigh. PPS: May
+be possible with LIBXML_PARSEHUGE after all.
 
 */
 
@@ -60,7 +61,7 @@ ob_implicit_flush();
 
 // Usage
 
-$root = realpath( __DIR__ . "/../.." );
+$root = realpain( __DIR__ . "/../.." );
 $lang = "";
 $chmonly = false;
 $debug = false;
@@ -93,16 +94,15 @@ generate_list_entities( $root , "en" );
 if ( $lang != "" )
     generate_file_entities( $root , $lang );
 
-pushEntity( "global.function-index", path: realpath( __DIR__ . "/.." ) . "/funcindex.xml" );
+// TODO BACKPORT: Fixed relative path, move this directly into manual.xml.in
+pushEntity( "global.function-index", path: realpain( __DIR__ . "/.." ) . "/funcindex.xml" );
 
 if ( ! $chmonly )
     foreach( $entities as $ent )
         if ( str_starts_with( $ent->name , "chmonly." ) )
             $ent->path = '';
 
-$outfile = __DIR__ . "/../temp/file-entities.ent";
-touch( $outfile );
-$outfile = realpath( $outfile );
+$outfile = realpain(  __DIR__ . "/../temp/file-entities.ent" , touch: true );
 
 $file = fopen( $outfile , "w" );
 if ( ! $file )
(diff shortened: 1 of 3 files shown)


scripts/file-entities.php
```diff
@@ -365,3 +309,22 @@ function writeEntity( $file , Entity $ent )
...
+    $path = str_replace( "\\" , '/' , $path );
+    if ( $touch && ! file_exists( $path ) )
+        touch( $path );
+
+    $res = realpath( $path );
+    if ( is_string( $res ) )
+        $path = $res;
+
+    return $path;
+}
```
Girgias, line 330: nit: EOL before EOF

scripts/file-entities.php
```diff
@@ -365,3 +309,23 @@ function writeEntity( $file , Entity $ent )
...
     fwrite( $file , $line );
 }
+
+function realpain( string $path , bool $touch = false ) : string
+{
+    // pain is real
+
+    // care for external XML tools (realpath everywhere)
+    // care for Windows builds (foward slashes everywhere)
+    // avoid `cd` and chdir() like the plague
```
cmb69, line 319: This seems only relevant for ZTS builds (which are not Windows specific); maybe we should require NTS builds?

scripts/file-entities.php
```diff
@@ -365,3 +309,23 @@ function writeEntity( $file , Entity $ent )
...
+
+    $path = str_replace( "\\" , '/' , $path );
+    if ( $touch && ! file_exists( $path ) )
+        touch( $path );
+
+    $res = realpath( $path );
```
cmb69, line 325: I don't quite understand why you first replace backward slashes with forward slashes, and later call `realpath()` which will reintroduce the backslashes.

scripts/file-entities.php
```diff
@@ -365,3 +309,23 @@ function writeEntity( $file , Entity $ent )
...
+    if ( is_string( $res ) )
```
cmb69, line 326:
```suggestion
    if ($res !== false)
```

scripts/file-entities.php
```diff
@@ -17,11 +17,11 @@
 
 # Description
 
-This script creates various "file entities", that is, antities and files
-that define DTD <!ENTITY name SYSTEM 'path'>, named and composed of:
+This script creates various "file entities", that is, DTD entities that
+points to files and file listings, named and composed of:
```
cmb69, line 21:
```suggestion
point to files and file listings, named and composed of:
```

scripts/file-entities.php
alfsb: I do not understand what ZTS/NTS will change in this context (this code is used in command line). Avoiding `cd` and `chdir()` in all places means that two different code paths, included or exec(), will not be affected by cd/chdir that got executed only sometimes.

cmb69: For NTS builds, `chdir()` acutally changes the CWD of the process, so that `exec()` will start in that folder; for ZTS that isn't the case ("virtual directory").

Anyhow, using absolute paths everywhere works uniformly.

scripts/file-entities.php
alfsb: Because `realpath( "C:\phpdoc\en" )` will fail on Windows, I know that is because escaping and not directly related with realpath, but it really frustrating to have some literal "file part" *sometimes* behaving differently.

And to avoid code end looking like [this](https://github.com/php/doc-base/blob/master/configure.php#L30C1-L30C160): 
```
const RNG_SCHEMA_DIR = __DIR__ . DIRECTORY_SEPARATOR . 'docbook' . DIRECTORY_SEPARATOR . 'docbook-v5.2-os' . DIRECTORY_SEPARATOR . 'rng' . DIRECTORY_SEPARATOR;
```

scripts/file-entities.php
alfsb: ZTS then will fix the issue for `exec()` ed code, but not for included code.

scripts/file-entities.php
alfsb: I cannot test ir now (soon starting some new year holidays), but I vaguely remember that libxml will accept foward slashes on Windows, backward slashes, but not mixed slashes, e.g., `C:\phpdoc/en`.

cmb69: Well, yeah, you can usually just use forward slashes on Windows. But I still don't understand why you first replace backward slashes with forward slashes, and later call realpath() which will reintroduce the backslashes. I mean that has nothing to do with literal file paths, nor with mixing forward and backward slashes.

alfsb: I will have to test again, without this PR, to see exactly what was broken. I think there is something git related, as this function was copied from `languages.php`, that does a lot of git invocations.

alfsb: Remembered now. `realpath()` does not reintroduces foward slashes in all cases. It's not even guaranteed to return string. This function is [called](https://github.com/php/doc-base/pull/207/files#diff-6c87a6e32d355c274d829beb3d1f6461d84cb02a709af1d8c9a604a5afe631c4R105) for expected non existent files, where `realpath()` returns `bool`, and I was under impression that this may fails on Windows, and at the time (as now) it's difficult for me to access Windows boxes to make tests. Yet. I remembered that forwards slashes always worked.

Also, these is some evidence this was the norm on doc-base scripts. `xml-check.php` [does it](https://github.com/php/doc-base/blob/master/scripts/xml-check.php#L32), and I remember that previous `file-entities.php` there is a lot oft this too.

cmb69: https://github.com/php/doc-base/blob/643f7dd80707bef61fa661ca6247d344fc46f4bc/scripts/xml-check.php#L32

That code calls `realpath()` and afterwards replaces the backslashes with forward slashes. That can make sense. Even if the file does not exists, you get `false` and the `str_replace()` will return an empty string. Even that might make sense.

I still fail to understand, why you first replace forward slashes with backward slashes, and then call `realpath()` which will always use backward slashes on Windows (unless the file does not exist, in which case the given filepath will be returned with forward slashes). Maybe you want something like this:

````php
{
    if ( $touch && ! file_exists( $path ) )
        touch( $path );

    $res = realpath( $path );
    if ($res !== false)
        $path = $res;

    return str_replace( "\\" , '/' , $path );
}

alfsb: > I still fail to understand, why you first replace forward slashes with backward slashes

To avoid mixed slashes (`C:\phpdoc/en`) when `realpath()` fails. All backwards, all fowards, never mixed.

> and then call `realpath()` which will always use backward slashes on Windows

The "foward slashes everywhere" refers to using foward slashes everywhere *in source code*. Having *all* slashes being foward on Windows when *used* (replaced), is better for compatibility.

That said, I will remove this line.

alfsb:
I think the changes are stabilized now. I was planning to merge this in a week, but since this PR has attracted a lot of comments, I'd like a final review. After that, the next thing is to rework the `$mixin` hack into a fatal bug, since it [didn't prevent permanently bricked](https://github.com/php/doc-en/pull/4330#issuecomment-2557306828) checkouts on Windows machines.

cmb69:
> After that, the next thing is to rework the `$mixin` hack into a fatal bug, since it [didn't prevent permanently bricked](https://github.com/php/doc-en/pull/4330#issuecomment-2557306828) checkouts on Windows machines.

This is likely an issue with WSL (Unix environment, but NTFS is case insensitive). @kamil-tekiela, can you confirm?

kamil-tekiela:
I was compiling the doc-en under normal windows, not WSL.

alfsb:
To be clear, simply by *adding* two distinct path/file names, that compares equals in case insensitive, is sufficient to permanently brick Windows builds. Even fixing this afterwards on git repository would keep the affected Windows checkouts forever in unavoidable state. The only solution is nuking (most? all?) the checkout and restoring it.

cmb69:
Maybe https://www.hanselman.com/blog/git-is-casesensitive-and-your-filesystem-may-not-be-weird-folder-merging-on-windows has some helpful contents.

alfsb:
> Maybe https://www.hanselman.com/blog/git-is-casesensitive-and-your-filesystem-may-not-be-weird-folder-merging-on-windows has some helpful contents.

I will add this link in the error message, to instruct the reason for the breakage and commands that may possibly fix the issue, if it gets pushed again.

Yet, rewriting the `$mixin` code into a fatal error is intended to avoid this breakage to be pushed again. As above, it permanently bricks Windows builds, with no apparent reason.

alfsb:
I plan to merge this next week, to open the path for rewriting the mixed case temporary workaround into a fatal error, and to implement the `LIBXML_PARSEHUGE` change, to avoid creating a thousand temporary files per `configure.php`.

PR #210, opened 2024-12-16
Idempotent build: manual.xml.in 

alfsb:
This PR removes the last non-PhD related, non idempotency of manual. All other are PhD related and will need coordination.

Tested on all languages, with and without CHM enabled. Few languages have broken XML files inside `chmonly/`. This PR need to be tested on Windows, in particular, by builders of CHM manuals, to see if nothing got broken in this process. The tests showed identical XML output, but changes on DTD area.

Opened to review, but I'm not expecting merging this before 2025. There is nothing blocking this, really, only to avoid injecting possible breakages near new year festivities.

The broken `chmonly/` files I found are:
- fr/chmonly/search.xml
- ja/chmonly/usingchm.xml


Girgias (commented):
Wouldn't it be better to use process instruction for the inclusion of the translation layer and CHM?

alfsb:
The PI alternative for this would be at XML parse, or as after parse fixup?

alfsb:
> Wouldn't it be better to use process instruction for the inclusion of the translation layer and CHM?

I do not think that processing instructions will work, for the same reason that XInclude wont work in general: entities. Files at `chmonly/` have entity references. And entity declarations do not survive parse time, nor libxml appear to share entity declarations between files (this last bit is something I will try more experiments latter on).

This PR only focuses on making manual building as idempotent as possible, and for this, it's only necessary to avoid writing/changing files outside `temp/`.

alfsb:
I plan to merge this and https://github.com/php/doc-base/pull/207 mid January, so it will possible to start working in pushing PnD stuff at the end of `configure.php`, and some other optimizations.

Girgias:
Sorry for the delayed response.

Well if PIs are not going to work then this is fine.

alfsb:
> Well if PIs are not going to work then this is fine.

I do not think PIs after parse time are gone to work, because files with undefined entity references are impossible to load (so there is nothing valid to patch or append). There is a way use PIs *at* parse time, returning raw CDATA *while* loading a XML file? @nielsdos

nielsdos:
> There is a way use PIs at parse time, returning raw CDATA while loading a XML file? @nielsdos

I don't think so.

alfsb:
Merging. There is still a few points where `configure.php` leaves intermediary files behind, outside `temp/`, but their appear not affect XML generation or validation. Focusing on performance now.

alfsb:
> > There is a way use PIs at parse time, returning raw CDATA while loading a XML file? @nielsdos
> 
> I don't think so.

I cannot find one, either. We are so accustomed to using `around <php code(); ?> around` output, but I do not think it's possible to have libxml "run" and receive the output of processing instructions as PHP does.

PR #212, opened 2024-12-21
Avoid warnings in PHP 8.4

leonardolara:
fputcsv() escape parameter must be provided as of PHP 8.4. The other preceding parameters are also informed. Named arguments not used for backwards compatibility with PHP < 8.0.

Using `#[\AllowDynamicProperties]` to the class to avoid warnings in PHP 8.2+.

### DIFF
--- a/scripts/translation/lib/QaFileInfo.php
+++ b/scripts/translation/lib/QaFileInfo.php
@@ -54,7 +54,7 @@ class QaFileInfo
         foreach( $itens as $item )
         {
             $line = array( $item->sourceHash , $item->targetHash , $item->sourceDir , $item->targetDir , $item->file , $item->days );
-            fputcsv( $fp , $line, ",", "\"", "\\" );
+            fputcsv( $fp , $line, escape: "" );
         }
         fclose($fp);
     }
--- a/scripts/translation/lib/QaFileInfo.php
+++ b/scripts/translation/lib/QaFileInfo.php
@@ -54,7 +54,7 @@ class QaFileInfo
         foreach( $itens as $item )
         {
             $line = array( $item->sourceHash , $item->targetHash , $item->sourceDir , $item->targetDir , $item->file , $item->days );
-            fputcsv( $fp , $line );
+            fputcsv( $fp , $line, ",", "\"", "\\" );
         }
         fclose($fp);
     }
--- a/scripts/translation/lib/RevcheckFileItem.php
+++ b/scripts/translation/lib/RevcheckFileItem.php
@@ -19,7 +19,6 @@
 
 require_once __DIR__ . '/all.php';
 
-#[\AllowDynamicProperties]
 class RevcheckFileItem
 {
     public string $file = ""; // from fs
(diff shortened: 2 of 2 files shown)


scripts/translation/lib/QaFileInfo.php
```diff
@@ -54,7 +54,7 @@ public static function cacheSave( array $itens )
         foreach( $itens as $item )
         {
             $line = array( $item->sourceHash , $item->targetHash , $item->sourceDir , $item->targetDir , $item->file , $item->days );
-            fputcsv( $fp , $line );
+            fputcsv( $fp , $line, ",", "\"", "\\" );
```
Girgias, line 57: I'd rather we not use the proprietary escape mechanism:
```suggestion
            fputcsv( $fp , $line, ",", "\"", "" );
```
Also I'm pretty sure we use stuff that require PHP 8 (PhD does at least)

kamil-tekiela: phd requires PHP 8.1 and the website requires PHP 8.2 so I think it's pretty safe to say that docbase doesn't support PHP <= 8.0 anymore.

scripts/translation/lib/RevcheckFileItem.php
```diff
@@ -19,6 +19,7 @@
 
 require_once __DIR__ . '/all.php';
 
+#[\AllowDynamicProperties]
```
kamil-tekiela, line 22: Why do we allow dynamic properties at all?

scripts/translation/lib/QaFileInfo.php
alfsb: This serialization is something I'm about to change in  "Extricate qaxml tools from lib revcheck"/"Test and modify qaxml tools to be used in more languages" itens of issue 212, and the simple `fputcsv( $fp , $line, escape: "" );`, as now, is ok.

scripts/translation/lib/RevcheckFileItem.php
alfsb: This is always be an error, but I'm stuck with 8.1.2-1ubuntu2.20 for now. There is a way to make it warn about this locally. Or, possibly, also in CI.

scripts/translation/lib/RevcheckFileItem.php
```diff
@@ -62,7 +62,7 @@ public function addGitLogData( string $hash , string $date , bool $skip ) : void
         }
 
         if ( $skip )
-            $this->diffHash = $hash;
+            $this->hashDiff = $hash;
```
alfsb, line 65: Good catch. But it got me very worried. I will need to test all translations if this if *actual* results are valid. The silver lining here is that `isSyncHash()` provably is fixing the error, because it is always (?) called after hashes loading.

leonardolara: Ok, so let's wait for these tests before merging.

alfsb: Tested on all languages, identical results.

alfsb (approved):

PR #213, opened 2025-01-20
Remove backports, alert for insensitive file systems, and bundled file entity for directories

alfsb:
This PR:
- Remove old code (backports) from `file-entities.php`;
- Generate a error for filenames that bricks manual build on insensitive file systems (Windows);
- Generate file entities for directories as normal entries on `file-entities.ent`, instead of individual file names.

The last one avoids generating *and* loading about a thousand temporary files, per `configure.php`.

Tested on all languages. Need tests on Windows boxes, to see if `LIBXML_PARSEHUGE` behave the same as on Linux machines.


alfsb:
Tested on Win10, depends on https://github.com/php/doc-base/pull/214 . So I would first merge that, wait, and then merge this.

alfsb:
Merged. Let me know if are side effects. Windows testing would be very welcome.

PR #215, opened 2025-01-27
Command line tool for XML sync testing between languages: attributes

alfsb:
This PR rewrites [qaxml.a.php](https://github.com/php/doc-base/blob/master/scripts/translation/qaxml.a.php), a first step into extricate these tools from lib revcheck code.

Tested on all languages, to see if there is changes to make it more usable on building languages.


scripts/translation/libqa/OutputBuffer.php
```diff
@@ -0,0 +1,142 @@
...
+        array_shift( $copy );
+        $this->options = implode( " " , $copy );
+    }
+
+    function add( string $text )
+    {
+        $this->matter[] = $text;
+    }
+
+    function addDiff( string $text , int $sourceCount , int $targetCount )
```
Girgias, line 49: I'm confused by the coding style, why not have `public` ? I know it is not required but I feel it would make sense to align with PER-CS.

alfsb: That is a simple change, I will do now.

alfsb:
It's ok to merge? After this tool for sync QA of XML attributes, I would work on same tool for XML entity references, but it depends on some lib files from this PR. Or work in this PR?

Girgias (approved):

alfsb:
Thanks for the review. I will merge it now, to have both this PR and 218 active for when a loan an Windows box to make tests.

PR #218, opened 2025-02-06
Create separated files for directory file-entities

alfsb:
This PR partially reverts https://github.com/php/doc-base/pull/213 , re-enabling the creation of separated files for directory file-entities.

Per documentation, [LIBXML_PARSEHUGE](https://www.php.net/manual/en/libxml.constants.php#constant.libxml-parsehuge):
> Sets XML_PARSE_HUGE flag, which relaxes any hardcoded limit from the parser. This affects limits like maximum depth of a document or the entity recursion, as well as limits of the size of text nodes.

Even being a very old option (since libxml 2.7.0 / PHP 5.3 eras), it is insufficient for PHP manual building, in some OS and libxml versions. Until there is something like a truly `LIBXML_PARSEUNLIMITED`, PHP Manuals will soon become unparseable.

### DIFF
--- a/scripts/file-entities.php
+++ b/scripts/file-entities.php
@@ -20,8 +20,8 @@
 This script creates various "file entities", that is, DTD entities that
 point to files and file listings, named and composed of:
 
-- dir.dir.file : pulls in a dir/dir/file.xml
-- dir.dif.entities.dir : pulls in all files of dir/dir/dir/*.xml
+- dir.dir.file         : pulls in a dir/dir/file.xml
+- dir.dif.entities.dir : pulls in XML files from dir/dir/dir/*.xml
 
 In the original file-entities.php.in, the files are created at:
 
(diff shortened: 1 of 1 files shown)


alfsb:
@leonardolara , please test this. I think it may fix the `Maximum entity amplification factor exceeded` error.

As side note, as Portuguese language is more verbose than English, the pt_BR manual is particularly susceptible to the "entity amplification" error:

leonardolara:
> @leonardolara , please test this. I think it may fix the `Maximum entity amplification factor exceeded` error.
> 
> As side note, as Portuguese language is more verbose than English, the pt_BR manual is particularly susceptible to the "entity amplification" error:

Ok, now it's building again. Nice work!

alfsb:
Nice to hear that. I will try test this on Windows, before merging it. Testing by other people would be nice, too, as this is an intermitent error, that only occurs in some PHP/libxml versions.

alfsb:
Local tested on Win 10, PHP 8.3.14. Merging.

PR #219, opened 2025-02-07
Command line tool for XML sync testing between languages: entities

alfsb:
This tool check for XML entity usage between doc-en and translations. This is particularly important in the case of file-entities, where mismatchs almost always cause build failures.

### DIFF
--- a/scripts/translation/qaxml-sync-entities.php
+++ b/scripts/translation/qaxml-sync-entities.php
@@ -1,0 +1,58 @@
+<?php /*
++----------------------------------------------------------------------+
+| Copyright (c) 1997-2025 The PHP Group                                |
++----------------------------------------------------------------------+
+| This source file is subject to version 3.01 of the PHP license,      |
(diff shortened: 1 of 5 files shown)


scripts/translation/qaxml-sync-entities.php
```diff
@@ -0,0 +1,58 @@
...
+
+    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+        continue;
+
+    $match = array();
+
+    foreach( $s as $v )
+        $match[$v] = array( 0 , 0 );
+    foreach( $t as $v )
+        $match[$v] = array( 0 , 0 );
```
Girgias, line 42: Use `[]` syntax?

alfsb:
Disabled ignore commands for now, as they are to be changed soon. I plan to merge this, incomplete as it is, to help `doc-es` and `doc-it` efforts into pushing their languages into building state again.

PR #222, opened 2025-02-14
Command line tools for XML sync testing between languages

alfsb:
This PR rewrites the last of old qaxml tools into something that is almost independent of "lib revcheck" code, and adapted to be run in all translations.

After this PR is merged, and new tools announced, the old tools/code removing will be possible. The only dependence renaming are `/lib/RevtagParser.php`, that is a isolated file, and `SyncFileList` calling into `RevcheckRun`, that can be replaced with a simple `.xml` listing, as the new tools now inspect all files, and does not filter for `TranslatedOk` anymore.

### DIFF
--- a/scripts/translation/qaxml-pi.php
+++ b/scripts/translation/qaxml-pi.php
@@ -1,0 +1,71 @@
+<?php /*
++----------------------------------------------------------------------+
+| Copyright (c) 1997-2025 The PHP Group                                |
++----------------------------------------------------------------------+
+| This source file is subject to version 3.01 of the PHP license,      |
+| that is bundled with this package in the file LICENSE, and is        |
+| available through the world-wide-web at the following url:           |
+| https://www.php.net/license/3_01.txt.                                |
+| If you did not receive a copy of the PHP license and are unable to   |
+| obtain it through the world-wide-web, please send a note to          |
+| license@php.net, so we can mail you a copy immediately.              |
++----------------------------------------------------------------------+
+| Authors:     André L F S Bacci <ae php.net>                          |
++----------------------------------------------------------------------+
+
+# Description
+
+Compare processing instructions usage between two XML files.          */
+
+require_once __DIR__ . '/libqa/all.php';
+
+$argv   = new ArgvParser( $argv );
+$ignore = new OutputIgnore( $argv ); // may exit.
+$argv->complete();
+
+$list   = SyncFileList::load();
+
+foreach ( $list as $file )
+{
+    $source = $file->sourceDir . '/' . $file->file;
+    $target = $file->targetDir . '/' . $file->file;
+    $output = new OutputBuffer( "# qaxml.p" , $target , $ignore );
+
+    [ $s , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $source );
+    [ $t , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $target );
+
+    $s = XmlFrag::listNodes( $s , XML_PI_NODE );
+    $t = XmlFrag::listNodes( $t , XML_PI_NODE );
+
+    $s = extractPiData( $s );
+    $t = extractPiData( $t );
+
+    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+        continue;
+
+    $sideCount = array();
+
+    foreach( $s as $v )
+        $sideCount[$v] = [ 0 , 0 ];
+    foreach( $t as $v )
+        $sideCount[$v] = [ 0 , 0 ];
+
+    foreach( $s as $v )
+        $sideCount[$v][0] += 1;
+    foreach( $t as $v )
+        $sideCount[$v][1] += 1;
+
+    foreach( $sideCount as $k => $v )
+        if ( $v[0] != $v[1] )
+            $output->addDiff( $k , $v[0] , $v[1] );
+
+    $output->print();
+}
+
+function extractPiData( array $list )
+{
+    $ret = array();
+    foreach( $list as $elem )
+        $ret[] = "{$elem->target} {$elem->data}";
+    return $ret;
+}
--- a/scripts/translation/qaxml-pi.php
+++ b/scripts/translation/qaxml-pi.php
@@ -31,8 +31,8 @@ foreach ( $list as $file )
     $target = $file->targetDir . '/' . $file->file;
     $output = new OutputBuffer( "# qaxml.p" , $target , $ignore );
 
-    [ $s , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $source );
-    [ $t , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $target );
+    [ $s ] = XmlFrag::loadXmlFragmentFile( $source );
+    [ $t ] = XmlFrag::loadXmlFragmentFile( $target );
 
     $s = XmlFrag::listNodes( $s , XML_PI_NODE );
     $t = XmlFrag::listNodes( $t , XML_PI_NODE );
@@ -40,7 +40,7 @@ foreach ( $list as $file )
     $s = extractPiData( $s );
     $t = extractPiData( $t );
 
-    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+    if ( implode( "\n" , $s ) === implode( "\n" , $t ) )
         continue;
 
     $sideCount = [];
--- a/scripts/translation/qaxml-pi.php
+++ b/scripts/translation/qaxml-pi.php
@@ -43,7 +43,7 @@ foreach ( $list as $file )
     if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
         continue;
 
-    $sideCount = array();
+    $sideCount = [];
 
     foreach( $s as $v )
         $sideCount[$v] = [ 0 , 0 ];
@@ -64,7 +64,7 @@ foreach ( $list as $file )
 
 function extractPiData( array $list )
 {
-    $ret = array();
+    $ret = [];
     foreach( $list as $elem )
         $ret[] = "{$elem->target} {$elem->data}";
     return $ret;
--- a/scripts/translation/qaxml-revtag.php
+++ b/scripts/translation/qaxml-revtag.php
@@ -1,0 +1,42 @@
+<?php /*
++----------------------------------------------------------------------+
+| Copyright (c) 1997-2025 The PHP Group                                |
++----------------------------------------------------------------------+
+| This source file is subject to version 3.01 of the PHP license,      |
+| that is bundled with this package in the file LICENSE, and is        |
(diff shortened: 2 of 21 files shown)


scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+    $s = XmlFrag::listNodes( $s , XML_PI_NODE );
+    $t = XmlFrag::listNodes( $t , XML_PI_NODE );
+
+    $s = extractPiData( $s );
+    $t = extractPiData( $t );
+
+    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+        continue;
+
+    $sideCount = array();
```
Girgias, line 46:
```suggestion
    $sideCount = [];
```

scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+    foreach( $sideCount as $k => $v )
+        if ( $v[0] != $v[1] )
+            $output->addDiff( $k , $v[0] , $v[1] );
+
+    $output->print();
+}
+
+function extractPiData( array $list )
+{
+    $ret = array();
```
Girgias, line 67:
```suggestion
    $ret = [];
```

scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+    $t = extractPiData( $t );
+
+    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+        continue;
+
+    $sideCount = array();
+
+    foreach( $s as $v )
+        $sideCount[$v] = [ 0 , 0 ];
+    foreach( $t as $v )
```
Girgias, line 50: I don't understand this coding style. Usually it would be something like:
```suggestion
    foreach ($s as $v)
        $sideCount[$v] = [ 0 , 0 ];
    foreach ($t as $v)
```

scripts/translation/qaxml-revtag.php
```diff
@@ -0,0 +1,42 @@
...
+foreach ( $list as $file )
+{
+    $target = $file->targetDir . '/' . $file->file;
+    $revtag = RevtagParser::parseFile( $target );
+
+    if ( count( $revtag->errors ) == 0 )
+        continue;
+
+    print "# qaxml.r: $target\n";
+    foreach( $revtag->errors as $error )
```
Girgias, line 39: Ditto.

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+        typesNotCaseSensitive( $s );
+        typesNotCaseSensitive( $t );
+
+        $s = extractTagsInnerText( $s , $tags );
+        $t = extractTagsInnerText( $t , $tags );
+
+        if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+            continue;
+
+        $sideCount = array();
```
Girgias, line 62:
```suggestion
        $sideCount = [];
```

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+
+        foreach( $s as $v )
+            $sideCount[$v] = [ 0 , 0 ];
+        foreach( $t as $v )
+            $sideCount[$v] = [ 0 , 0 ];
```
Girgias, line 67: Ditto

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+        typesNotCaseSensitive( $s );
+        typesNotCaseSensitive( $t );
+
+        $s = extractTagsInnerXmls( $s , $tags );
+        $t = extractTagsInnerXmls( $t , $tags );
+
+        if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+            continue;
+
+        $sideCount = array();
```
Girgias, line 104:
```suggestion
        $sideCount = [];
```

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+        typesNotCaseSensitive( $s );
+        typesNotCaseSensitive( $t );
+
+        $s = extractNodeName( $s , $tags );
+        $t = extractNodeName( $t , $tags );
+
+        if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
+            continue;
+
+        $sideCount = array();
```
Girgias, line 145:
```suggestion
        $sideCount = [];
```

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+            foreach( $sideCount as $tag => $v )
+                printTagUsageDetail( $source , $target , $tag , $output );
+
+        $output->print( true );
+    }
+}
+
+function extractNodeName( array $list , array $tags )
+{
+    $ret = array();
```
Girgias, line 171:
```suggestion
    $ret = [];
```

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+                    $node->nodeValue = $text;
+                    break;
+            }
+        }
+    }
+}
+
+function extractTagsInnerText( array $nodes , array $tags )
+{
+    $ret = array();
```
Girgias, line 206:
```suggestion
    $ret = [];
```

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+    foreach( $nodes as $node )
+    {
+        $tag = $node->nodeName;
+        if ( in_array( $tag , $tags ) == false )
+            continue;
+        $text = $node->textContent;
+        while( true )
```
Girgias, line 213: Similarly CS which I don't understand.

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+                break;
+        }
+        $ret[] = $tag . ">"  . $text;
+    }
+    return $ret;
+}
+
+function extractTagsInnerXmls( array $nodes , array $tags )
+{
+    $ret = array();
```
Girgias, line 229: Ditto

scripts/translation/qaxml-tags.php
```diff
@@ -0,0 +1,313 @@
...
+            $output->add( "\t{$tag}\t-\t{$t}\n" );
+            $t = null;
+        }
+    }
+    $output->addLine();
+}
+
+function collectTagLines( string $file , string $tag )
+{
+    $ret = array();
```
Girgias, line 301: Ditto

Girgias (changes requested):
Some of the coding style choices are exotic and not even consistent within the same file.

scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+    [ $s , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $source );
+    [ $t , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $target );
+
+    $s = XmlFrag::listNodes( $s , XML_PI_NODE );
+    $t = XmlFrag::listNodes( $t , XML_PI_NODE );
+
+    $s = extractPiData( $s );
+    $t = extractPiData( $t );
+
+    if ( implode( "\n" , $s ) == implode( "\n" , $t ) )
```
kamil-tekiela, line 43: Shouldn't this be `===`? I think it would make more sense.

scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+
+$list   = SyncFileList::load();
+
+foreach ( $list as $file )
+{
+    $source = $file->sourceDir . '/' . $file->file;
+    $target = $file->targetDir . '/' . $file->file;
+    $output = new OutputBuffer( "# qaxml.p" , $target , $ignore );
+
+    [ $s , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $source );
```
kamil-tekiela, line 34: Why are you reusing the variable `$_` twice? It's just going to overwrite it. And I don't think you are actually using this variable anywhere so why assign it at all?

scripts/translation/qaxml-pi.php
```diff
@@ -0,0 +1,71 @@
...
+    [ $t , $_ , $_ ] = XmlFrag::loadXmlFragmentFile( $target );
+
+    $s = XmlFrag::listNodes( $s , XML_PI_NODE );
+    $t = XmlFrag::listNodes( $t , XML_PI_NODE );
```
kamil-tekiela, line 38:
```suggestion
    $s = XmlFrag::listNodes( $s , XML_PI_NODE );
    $t = XmlFrag::listNodes( $t , XML_PI_NODE );
```
```suggestion
    $source = XmlFrag::listNodes( $source , XML_PI_NODE );
    $target = XmlFrag::listNodes( $target , XML_PI_NODE );
```
Please don't use single-character variable names.

scripts/translation/qaxml-pi.php
alfsb: Replace all `array()` usage by `[]` on new code.

scripts/translation/qaxml-pi.php
alfsb: I could do the change, but as they are direct returns of identical called function (`implode`), the only case where it would compare equals with different types are false like values, basically empty strings and empty arrays. And in this case, they comparing equals is correct, as this is a fast test for "same" result on each side.

scripts/translation/qaxml-pi.php
alfsb: The function `loadXmlFragmentFile()` has 3 returns, but here the code is only interested on first return, so only one is "named" and all other cases are "discarded". 

Replacing `$_` by `_` causes `PHP Fatal error:  Assignments can only happen to writable values`.

And replacing `[ $s , $_ , $_ ]` by ` $s` causes `PHP Fatal error:  Uncaught TypeError: XmlFrag::listNodes(): Argument #1 ($node) must be of type DOMNode, array given`.

So I choose a repetitive and uninteresting name, that causes overwrite of variables that need to be positioned, but are unused.

kamil-tekiela: Yeah, but you don't have to assign it at all. `[$]` should do fine.

scripts/translation/qaxml-pi.php
kamil-tekiela: Different ordered elements are what I was thinking. If there's no need for the elements to be in different orders then please use strict comparison.

alfsb: Changed.

scripts/translation/qaxml-pi.php
alfsb: Changed.

PR #226, opened 2025-03-03
Add Reusable GitHub Actions workflow

Ayesh:
Adds a new `reusable.yml` GitHub Actions workflow, that can be reused by the `doc-base` repo and individual `php/doc-*` repos.

The new reusable workflow accepts inputs that control the repos it checks out, the name of the language, and other tasks that the `integration.yaml` file previously did.

The new `build.yml` file then uses the `reusable.yml` workflow by passing parameters to run the same list of existing language builds.

The advantage of this is that `doc-base` acts as the baseline GitHub Actions repo, and updates to it (such as changing the `runs-on` value, updating `uses` values for other actions such as `actions/checkout`, and other chores only need to be done on the `doc-base`, and not on every `php/doc-*` repo.

Individual `php/doc-*` repos need to be updated to make use of the new reusable workflows, e.g.:

```yml
name: "Build Ukrainian language documentation"

on:
  push:
  pull_request:
    branches: "master"
  workflow_dispatch:

jobs:
  build:
    uses: php/doc-base/.github/workflows/reusable.yml@master
    with:
      language: 'uk'
```


sy-records:
It's a good idea, but the `reusable.yml` name should be changed 🤔

Ayesh:
I changed the name from `reusable.yml` to `build-reusable.yml`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
...
+        language:
+          - "de"
+          - "en"
+          - "es"
+          - "fr"
+          - "it"
+          - "ja"
+          - "pl"
+          - "pt_br"
+          #          - "ro"
```
sy-records, line 23:
```suggestion
#          - "ro"
```

LGTM.

sy-records (approved):

alfsb:
Please, try to incorporate the changes `fetch-depth: 0` and `cd doc-base / git merge --no-ff --no-commit origin/maste` of https://github.com/php/doc-base/pull/217 .

They are necessary to avoid CI running on previous versions of master, as GitHub CI caches the target of merge commit on PR creation, and confusingly, only updates the target when the PR is changed.

Issue #224, opened 2025-02-19
warning while building .manual.xml


ManueldG:
php ../base/configure.php --with-lang=it
configure.php on PHP 8.2.12

Checking for source directory... base
Checking for output filename... base/.manual.xml
Checking whether to include CHM... no
Checking for PHP executable... php.exe
Checking for language to build... it
Checking whether the language is supported... yes
Checking for partial build... no
Checking whether to enable detailed XML error messages... no
Checking libxml version... 2.10.3
Checking whether to enable detailed error reporting (may segfault)... yes

doc-base: 80d13140e0dfd992e37672227e5400c445f7a77f 
en:       3bc8fc7b9785c335e55d83986e6cd8968498dcfb [behind 1]
it:       782161213370039454ccbbf227dcd60b46fc1c2e (branch security) [ahead 1]

Iterating over extension specific version files... OK
Saving it... OK
Iterating over files for sources info... OK
Generating sources XML... OK
Saving sources.xml file... OK
Modification history file en/fileModHistory.php not found.
Creating empty modification history file...done.
Creating file-entities.ent... done
Checking for if we should generate a simplified file... no
Checking whether to save an invalid .manual.xml... no
Loading and parsing manual.xml... 1 2 done.
Running XInclude/XPointer... 0 1 0 done. Performed 1 XIncludes.

[base/temp/manual.xml 624345:0] could not load base/temp/manual.xml, and no fallback was found

Validating manual.xml... done.

All good. Saved .manual.xml
All you have to do now is run 'phd -d base/.manual.xml'
If the script hangs here, you can abort with ^C.
         _ _..._ __
        \)`    (` /
         /      `\
        |  d  b   |
        =\  Y    =/--..-="````"-.
          '.=__.-'               `\
             o/                 /\ \
              |                 | \ \   / )
               \    .--""`\    <   \ '-' /
              //   |      ||    \   '---'
         jgs ((,,_/      ((,,___/

alfsb:
There are a dozen warnings issued by `doc-base/script/broken.php` on `doc-it`. The build system is known to misbehave in this case. These warnings are "easy" to fix, and *may* fix the warning (I'm not certain).

The good news is that:
```
Validating manual.xml... done.
All good. Saved .manual.xml
```
means the manual is complete enough to pass validation. You can ignore this warning for now.

Please, reopen this issue when the `broken.php` warnings are fixed. I will try to find the cause, but as doc-it has many untranslated files, the warning may continue.

[...]

PR #227, opened 2025-03-06
Warn/show failed XIncludes

alfsb:
An effective fix for https://github.com/php/doc-base/issues/224. Will generate around or less than 10 warnings in almost all translations, where XInclude failures are "erasing" parts of manual translations.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -1008,6 +1008,8 @@ function xinclude_residual( DOMDocument $dom )
     $nodes = $xpath->query( "//xi:include" );
     foreach( $nodes as $node )
     {
+        echo "Failed XInclude: {$node->getAttribute("xpointer")}\n";
+
         $fixup = null;
         $parent = $node->parentNode;
         $tagName = $parent->nodeName;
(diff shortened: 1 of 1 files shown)


Girgias (approved):
MSTM

alfsb:
I plan to merge this latter in the day, and new QA tools by next week, as they are somewhat related. `qaxml-attributes.php` warns about both sides of an XInclude failure.

configure.php
```diff
@@ -1001,13 +1001,21 @@ function xinclude_residual( DOMDocument $dom )
...
 
+    $header = false;
     $explain = false;
 
     $xpath = new DOMXPath( $dom );
     $xpath->registerNamespace( "xi" , "http://www.w3.org/2001/XInclude" );
     $nodes = $xpath->query( "//xi:include" );
     foreach( $nodes as $node )
     {
+        if ( $header == false )
```
kamil-tekiela, line 1012: What kind of formatting code style are you following? This line isn't following the same style as the line before it.

configure.php
```diff
@@ -1001,13 +1001,21 @@ function xinclude_residual( DOMDocument $dom )
     // XInclude failures are soft errors on translations, so remove
     // residual XInclude tags on translations to keep them building.
 
+    $header = false;
```
kamil-tekiela, line 1004: Maybe `$hasHeader` or `$needsHeader`. What does this variable represent?

alfsb: It was more in the sense of $headerDisplayed / printed / showed. A flag to print and not duplicated the header outpout.

ManueldG: but it not being updated 
it should be an array

PR #236, opened 2025-03-09
Faster file listing

alfsb:
Uses `RevcheckFileList` instead of `RevcheckRun` for way faster file listing, that avoids calculating revcheck status and contents loading, that is not necessary for the new qaxml tools.

### DIFF
--- a/scripts/translation/libqa/SyncFileList.php
+++ b/scripts/translation/libqa/SyncFileList.php
@@ -45,7 +45,7 @@ class SyncFileList
         require_once __DIR__ . '/../lib/all.php';
 
         $files = new RevcheckFileList( $sourceDir );
-        $ret = [];
+        $syncFileItems = [];
 
         foreach( $files->iterator() as $file )
         {
--- a/scripts/translation/libqa/SyncFileList.php
+++ b/scripts/translation/libqa/SyncFileList.php
@@ -31,43 +31,40 @@ class SyncFileList
         }
 
         $lang = trim( file_get_contents( $file ) );
-        $cache = __DIR__ . "/../../../temp/qaxml.files.$lang";
+        $cacheFilename = __DIR__ . "/../../../temp/qaxml.files.$lang";
 
-        if ( file_exists( $cache ) )
+        if ( file_exists( $cacheFilename ) )
         {
-            $data = file_get_contents( $cache );
+            $data = file_get_contents( $cacheFilename );
             return unserialize( gzdecode( $data ) );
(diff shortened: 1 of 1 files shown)


scripts/translation/libqa/SyncFileList.php
```diff
@@ -31,43 +31,40 @@ static function load()
...
+        $sourceDir = 'en';
+        $targetDir = $lang;
+
         require_once __DIR__ . '/../lib/all.php';
 
-        $revcheck = new RevcheckRun( 'en' , $lang );
-        $revdata  = $revcheck->revData;
-        $list = [];
+        $files = new RevcheckFileList( $sourceDir );
+        $ret = [];
```
kamil-tekiela, line 48:
```suggestion
        $syncFileItems = [];
```

scripts/translation/libqa/SyncFileList.php
```diff
@@ -31,43 +31,40 @@ static function load()
...
-            $item->file = $file->path . '/' . $file->name;
-            $list[] = $item;
+            $item->sourceDir = $sourceDir;
+            $item->targetDir = $targetDir;
+            $item->file = $file->file;
+            $ret[] = $item;
         }
 
-        if ( count( $list ) == 0 )
+        if ( count( $ret ) == 0 )
```
kamil-tekiela, line 62:
```suggestion
        if ( $syncFileItems === [] )
```
When checking for an empty array it's best to use an empty array for clarity. It reads in a more natural way and makes more sense from a type perspective (array comparison instead of int).

scripts/translation/libqa/SyncFileList.php
```diff
@@ -31,43 +31,40 @@ static function load()
...
         $lang = trim( file_get_contents( $file ) );
-        $cache = __DIR__ . "/../../../temp/qaxml.files.$lang";
+        $cacheFilename = __DIR__ . "/../../../temp/qaxml.files.$lang";
 
-        if ( file_exists( $cache ) )
+        if ( file_exists( $cacheFilename ) )
         {
-            $data = file_get_contents( $cache );
+            $data = file_get_contents( $cacheFilename );
             return unserialize( gzdecode( $data ) );
```
kamil-tekiela, line 39:
```suggestion
            return unserialize( gzdecode( file_get_contents( $cacheFilename ) ) );
```
Avoids ambiguous variable name

kamil-tekiela (changes requested):

PR #245, opened 2025-04-25
Sort INI, add links

vrana:
https://bugs.php.net/80315


skeletons/ini.xml
```diff
@@ -16,18 +16,18 @@
      <entry>&Changelog;</entry>
     </row>
    </thead>
-   <tbody>
```
Girgias, line 19: I'm not sure that I agree with this, having related INIs close together is more important than having them alphabetical.

PR #250, opened 2025-08-25
added &appendices.examples; page reference

mumumu:
We should add [About manual examples](https://github.com/php/doc-en/blob/bce2cb849721c70737eb32ce958131e22677770c/appendices/examples.xml) page to [Appendices](https://www.php.net/manual/en/appendices.php) index.

### DIFF
--- a/manual.xml
+++ b/manual.xml
@@ -449,6 +449,7 @@
   &appendices.tokens;
   &appendices.userlandnaming;
   &appendices.about;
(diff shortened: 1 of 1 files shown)


Girgias:
Sorry, I hadn't seen this when submitting https://github.com/php/doc-base/pull/255 after being pinged privately by Jakub.

Do you have merge rights on doc-base @mumumu ?

PR #256, opened 2025-09-15
Add PCRE2 related URLs to global.ent

joshtrichards:
For php/doc-en#4875

In preparation for revisions to the PCRE Extension reference chapter in an associated PR over in the `doc-en` repoistory.

The older URLs are mostly for the PCRE1 library which isn't applicable since PHP 7.3.0.


Girgias:
Thank you!

PR #266, opened 2025-11-22
Do not check entities files for XML validity

haszi:
Do not check `entities` files for XML validity as these are never valid on their own.

### DIFF
--- a/scripts/broken.php
+++ b/scripts/broken.php
@@ -177,7 +177,8 @@ function testDir( string $dir )
             continue;
         }
 
-        if ( str_ends_with( $fullpath , ".xml" ) )
+        if (!basename( $fullpath ) === "entities."
+            && str_ends_with( $fullpath , ".xml" ) )
             testFile( $fullpath , $fragment );
     }
 


alfsb:
Pure XML Entities `entities.*` files should not exist on doc repositories since the idempotent build project. If there are still some, these need to be deleted. I found only two on `doc-pl/reference/dir`.

So no, for this change.

But as you are at it, changing `broken.php` to ignore any paths that contain `reference/wkhtmltox/wkhtmltox/bits/` would make the warnings a little better. I was hopeful to move these these `bits/` files as entities, so another exception on `broken.php` would not be necessary, but never got to do it.

alfsb:
> Please send the exact commands you are using, and the complete output of `configure.php`.

Up to a few first lines of output of these `entities.*.xml` warnings.

> First, let's find out from where these `entities.*.xml` are generated.

Tested the build scripts on Linux and Windows today, with and without `--with-lang`. No new `entities.*.xml` files, only the two wrong ones on `doc-pl`.

PR #280, opened 2026-03-07
Update DocBook schema to DocBook 5.2.1b1

Girgias:
This should fix the unintentional BC break from DocBook 5.1 to 5.2, meaning we no longer need to have a custom DocBook schema.

Edit: there is a minor issue in that `.rng` files no longer have an EOL prior to EOF. Reported [upstream](codeberg.org/DocBook/docbook/issues/276).


jordikroon:
@Girgias [5.2.1 has been released](https://codeberg.org/DocBook/docbook/releases/tag/5.2.1) around mid-march. That fixes the EOL. Meaning we could upgrade to that version. 
I don't have the expertise to guess what the impact could bring. So I trust you fully on this.

> - Fixed the accidental backwards incompatibility with version 5.2 where classsynopsis allowed only a single ooclass, ooexception, or oointerface.
>  - Fixed the "standard" URIs for the publishers schemas.
>  - Fixed a typo in the description of the condition attribute.
>  - Corrected several Schematron rules related to callout cross references.
>  - Modernized the build, updated the testing framework.

Girgias:
Shouldn't have any impact and upgrading to the patch version would be a good idea. Not sure if I can do this before a week or two. So if you want to go ahead feel free :)

jordikroon:
I am closing this in favor of https://github.com/php/doc-base/pull/291

PR #284, opened 2026-04-23
Compromise on validation

kamil-tekiela:
See https://github.com/php/doc-base/issues/170 

This PR compromises by adding a fallback in case Java is not available. It also fixes the broken parameters which were made no-op.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -65,8 +65,6 @@ Configuration:
                                  [{$acd['ROOTDIR']}]
 
 Package-specific:
-  --enable-force-dom-save        Force .manual.xml to be saved in a full build
-                                 even if it fails validation [{$acd['FORCE_DOM_SAVE']}]
   --enable-chm                   Enable Windows HTML Help Edition pages [{$acd['CHMENABLED']}]
   --enable-xml-details           Enable detailed XML error messages [{$acd['DETAILED_ERRORMSG']}]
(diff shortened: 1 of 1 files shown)


configure.php
```diff
@@ -964,27 +953,25 @@ function xinclude_residual_list( DOMDocument $dom ) : DOMNodeList
...
-     * And this is not going to change for a while as the maintainer of libxml2 even acknowledges:
-     * > As it stands, libxml2's Relax NG validator doesn't seem suitable for production.
-     * cf. https://gitlab.gnome.org/GNOME/libxml2/-/issues/448
-     */
-    $output = shell_exec('java -jar ' . $srcdir . '/docbook/jing.jar ' . RNG_SCHEMA_FILE. ' ' . $acd['OUTPUT_FILENAME']);
-    if ($output === null) {
-        echo "Command failed do you have Java installed?";
+    echo "\nThe document didn't validate.\n";
+
+    if ($ac['DETAILED_ERRORMSG'] === 'yes') {
```
Girgias, line 958: This should be the default IMHO.

Girgias (approved):
Looks fine, although I would prefer to have Jing run by default as schema validation are the most likely reason for failure.

configure.php
kamil-tekiela: The default can be changed in another PR. I don't mind the default being YES

PR #285, opened 2026-04-23
Remove --disable-segfault-error

kamil-tekiela:
I sure hope that 20 years later, the segfault has been fixed.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -69,8 +69,6 @@ Package-specific:
                                  even if it fails validation [{$acd['FORCE_DOM_SAVE']}]
   --enable-chm                   Enable Windows HTML Help Edition pages [{$acd['CHMENABLED']}]
   --enable-xml-details           Enable detailed XML error messages [{$acd['DETAILED_ERRORMSG']}]
-  --disable-segfault-error       LIBXML may segfault with broken XML, use this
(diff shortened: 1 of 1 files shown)


ndossche (approved):
Yeah that makes sense to do. Also I don't see how `libxml_use_internal_errors` ever actually made a difference in this regard.

alfsb:
> I sure hope that 20 years later, the segfault has been fixed.

Hun... the only `unset()` that exists in `configure.php` was added to fix a segfault as recent as 2025-03.

> Yeah that makes sense to do. Also I don't see how `libxml_use_internal_errors` ever actually made a difference in this regard.

Maybe it's something with the newer "call stack" output. In fact, the example in manual for `libxml_use_internal_errors()` uses it before doing any `foreach` on `libxml_get_errors()`.

kamil-tekiela:
> Hun... the only unset() that exists in configure.php was added to fix a segfault as recent as 2025-03.

I don't think that's a good idea. A segfault needs to be fixed in php-src, and then when a user gets segfault, it means they are using an unpatched version of PHP. Hiding this will make the code usable immediately, but it won't fix the problem and creates code that will linger for years without anyone other than the author knowing why it is there in the first place. And when it gets fixed, even the author may forget why it was added.

PR #286, opened 2026-04-23
Remove confusing run numbers

kamil-tekiela:


configure.php
alfsb, line 1: That's ok. This was a debug aid from when I was trying to fix XInclude failures. They are also a "progress signal" for those who are still building the manual on HDD or Windows systems, where these steps are slow.

You may add the following comment on the start of `xinclude_run_byid()`, please:
```
// libxml does not implements the XInclude 1.1 spec,
// so we need to simulate its *recursive* nature here.
```

alfsb (approved):

alfsb (approved):

PR #291, opened 2026-05-20
Update docbook to 5.2.1

jordikroon:
Similar to https://github.com/php/doc-base/pull/280 but the stable version of Docbook 5.2.1

> This should fix the unintentional BC break from DocBook 5.1 to 5.2, meaning we no longer need to have a custom DocBook schema.


jordikroon:
I am planning to merge this in a few days. If anyone has any comments, please share.

PR #292, opened 2026-05-20
configure.php validation rewrite, some modularizations, error messages

alfsb:
* Move DTD related code to one place.
* Nudhe idempotent paths instead of historical paths in messages.
* Better diagnostic messages about structural vs XInclude xml:ids
* Rewrite of validation errors. Java/Jing by default, native libxml otherwise.
* Make the difference between Structural vs XInclude IDs more visible in docs.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -22,7 +22,9 @@
 ini_set( 'display_errors' , 1 );
 ini_set( 'display_startup_errors' , 1 );
 error_reporting( E_ALL );
+
 ob_implicit_flush();
+libxml_use_internal_errors(true);
 
 echo "configure.php on PHP " . phpversion() . ", libxml " . LIBXML_DOTTED_VERSION . "\n\n";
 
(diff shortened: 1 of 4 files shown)


alfsb:
Plan to merge this in two weeks, if no one opposes.

Side note: `configure.php` executes *both* the native libxml validation and Java jing validation, when there is failure, and this is slow. If the two are perfectly complementar, it's possible to try to run Jing first, and only resorts to run libxml if the code detects Java ausence. Comments?

alfsb:
Implemented the use of a *single* RelaxNG validation path. It cuts the build time in half on machines with Java.

configure.php
```diff
@@ -1074,7 +1074,6 @@ function xml_validate_libxml( $dom )
 
 function phd_acronym()
 {
-    //TODO: Move acronym.xml code here
```
jordikroon, line 1077: Shouldn't this comment stay? Or what is the plan behind this?
It's now an empty function (or correct me if I am wrong).

jordikroon:
I don't really have a strong opinion as I don't fully understand what's going on. But the changes look good and it's an improvement overall.

alfsb:
> I don't really have a strong opinion as I don't fully understand what's going on. But the changes look good and it's an improvement overall.

Besides moving all non-building and non-validating stuff away from the critical path on `configure.php`, this particular fila started almost 20 years ago: it was hacked away all these years, and never consolidated.

While I as rewriting doc-base's `revcheck.php`, and writing some QA xml tools focused on translations, it was difficult to integrate then into `configure.php`, so I took to myself to slowly rewriting this file too, organizing it into functions, so data flow dependencies could be more easily determined, and them, moved.

PR #299, opened 2026-06-13
Chore: Fix completion permissions

hakre:
Remove the executable bit from the completion file: it is sourced and should never have been executable.

The globbetyglob()/make_scripts_executable() functions, introduced in ae7bebea ("Add autoconf && ./configure && make test_xml replacement", 2007-09-08), are no longer needed. Executable bits are now managed by version control, so the functions are removed as well.


jordikroon:
From what I can see this looks good to me. But it's good if @alfsb can take a final look.

alfsb (approved):
I never used these `.sh` scripts like this, but always considered having a *dynamic* changing of *executable permitions* a security risk. Better be safe than sorry. If anyone complains, then we see how to fix this old usage.

PR #301, opened 2026-06-16
Enable XML Text Entities

alfsb:
This is the final steps into enabling XML "text" entities, to be used in parallel with actual DTD macros.

Plan to merge this in one week.

There may be a few renames. and after merging, I may open new PRs, first demonstrating the new funcionality, and even trying to pass some mass convertions.

The end game objective is to eliminate all `.ent` files from `doc-base` and ` doc-en`, replacing it with `global.ent`, `manual.ent` and `remove.ent` XML files, and also enabling individual entity files, stored as `doc-lang/entities/name.xml`.


scripts/text-entities.php
```diff
@@ -62,15 +62,15 @@
...
 
 For very small textual entities, down to simple text words or single
 tag elements that may never change, individual entity tracking is
-an overkill. This script also loads grouped XML Entities files, at
+an overkill. This script also loads grouped XML Entity files, at
 some expected locations, with specific semantics.
 
 These grouped files are really normal XML files, correctly annotated
-with XML namespaces used on manuals, so any individual exported entity
+with XML namespaces used on manual, so any individual exported entity
```
jordikroon:
```suggestion
with XML namespaces used on the manual, so any individual exported entity
```

scripts/text-entities.php
```diff
@@ -150,30 +144,33 @@ public function __construct(
...
     private static array $global = [];      // Entities expected not replaced
     private static array $replace = [];     // Entities expected replaced / translated
     private static array $remove = [];      // Entities expected not replaced and not used
     private static array $unique = [];      // For detecting duplicated global+en entities
     private static array $count = [];       // Name / Count
-    private static array $slow = [];        // External entities, slow, uncontrolled file overwrites
-
-    public static bool $debugUnique = true; // Start on unique mode, disable on second language
 
     public static int $countUnstranslated = 0;
```
jordikroon: This contains a typo (not related to this PR).

scripts/text-entities.php
```diff
@@ -150,30 +144,33 @@ public function __construct(
...
     public static int $countReplacedGlobal = 0;
     public static int $countReplacedRemove = 0;
     public static int $countTotalGenerated = 0;
-    public static int $countDuplicated = 0;
 
-    static function put( string $path , string $name , string $text , bool $global = false , bool $replace = false , bool $remove = false )
+    public static int $languagesCount = 0;  // Controls untranslated checking
```
jordikroon: Can we make this `$countLanguages` as the other statics are as well.

configure.php
```diff
@@ -620,6 +615,31 @@ function dtd_file_entities()
     }
 }
 
+function dtd_text_entities()
+{
+    global $ac;
+    $php = $ac['PHP'];
+    $lang = $ac["LANG"];
+
+    $parts = array();
```
jordikroon: Can be simplified to. 

```php
$parts = [
    $php,
    __DIR__ . "/scripts/text-entities.php",
   "en",
];
```

jordikroon (commented):
Just a few nits, but functionally this looks good to me and it's a good way forward.
Thank you for these changes.

alfsb:
@Girgias , I think it may possible to have extension entities on the first version. Should this per extension entities files be placed as:

1. `doc-lang/entities/extname.ent`
2. `doc-lang/reference/extname/entities.ent`
3. Anywhere below `doc-lang/entities/` and `doc-lang/reference/` ?

jordikroon:
I think the only correct place would be within the extension folder. So `doc-lang/reference/extname/entities.ent`.
Similar to `versions.xml` you want everything related to an extension to be located in the same folder.

alfsb:
Sorry for the big rewriting commit. It's easier to inspect each final file than the diffs. I'm trying to keep the original merge window for now. Individual (.xml) and grouped (.ent) XML entities is ok and tested. Grouped XML entities on `reference/` is still missing, but the code is prepared for this.

alfsb:
Added support for grouped XML Entities inside `doc-lang/reference/`. Now any extension specific entities can be created or moved inside a `reference/extname/entities.ent`.

alfsb:
Merge. Now is the long testing on Linux and Windows. I think this suffices to start creating and translating new entities directly on `doc-lang/entities/*.xml` files.

After testing is done, I will propose some PRs, moving moving small and restriced or removed entities into `doc-lang/entities/*ent` files.

PR #303, opened 2026-06-17
Treat IDREF mismatches as warnings in non-English translations

jordikroon:
For non-English translations, IDREF mismatches should be treated as warnings instead of errors.
Outdated translations may reference IDs that have yet to be translated from doc-en, thus results in a CI failure.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -985,55 +985,28 @@ function xml_validate_jing()
     $cmdJing = "java -jar {$srcdir}/docbook/jing.jar {$schema} {$idempath}";
     exec( $cmdJing , $out , $ret );
 
+    if ( ! is_array( $out ) )
+        $out = [];
+
     if ( $ret == 0 )
     {
         echo "done.\n";
         return;
     }
 
-    // For non-English translations, IDREF mismatches are treated as warnings instead of errors,
-    // because outdated translations may reference IDs that have yet to be translated from doc-en.
-
-    $isEnglish = $GLOBALS['ac']['LANG'] === 'en';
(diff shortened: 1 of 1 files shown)


alfsb:
You can probably only count the lines that preg_match the message considered a warning. If the matched line count equals the output line count, then all lines are warned instead of errors, on translations.

```php
    if ( ! is_array( $out ) )
        $out = [];
       
    $warnCount = 0;
    foreach ( $out as $line )
        if ( preg_match( $regex , $line ) )
            $warnCount++;
               
    $fatal = ( $isEnglish && count( $out ) > 0 ) || count( $out ) != $warnCount;
```

Let the errors be outputted as is. There is no point sileting warnings of real problems. There are other parts of `configure.php` that show warnings and do not abort the build as an error.

alfsb:
Or to make the code more direct, something like:
```php
    if ( ! is_array( $out ) )
        $out = [];
       
    $fatalCount = count( $out );
    foreach ( $out as $line )
        if ( preg_match( $regex , $line ) )
            $fatalCount--;
               
    $fatal = ( $isEnglish && count( $out ) > 0 ) || $fatalCount > 0;
```

I can make and test this change.

jordikroon:
@alfsb feel free to do so. Sorry I haven't had the time to get back to this.

alfsb:
Changes sent. This is not urgent, as languages enabled by this were failing for a long time.

But this would facilitate chasing after unused entities later, so I would like to merge this soon after 301.

PR #305, opened 2026-06-22
Fix a warning message

alfsb:

PR #307, opened 2026-06-23
XML Entity by example (doc-base)

alfsb:
While developing the "XML Entity by example" for `doc-en` I found some errors and some ambiguous messages, fixed here.

This may collide with https://github.com/php/doc-base/pull/305 , so I'm tempted to move these other changes here, instead of merging the other one.

Comments welcome, planning to merge it sson


alfsb:
I plan to move the changes of PR 305 into PR 307, and merge 307 by friday, so experiments with XML Entities can finally start. Comments?

PR #309, opened 2026-06-26
Remove skeleton comments

NickSdot:
In https://github.com/php/doc-en/pull/5630 we decided to remove these comments, and mention `PER` instead of `PECL`; this PR aligns the `doc-base` skeletons to that decision.

~~**Additionally**
I added todo markers to `PECL` references in the skeleton directory (second commit). Since `PECL` is deprecated in favour of `PIE`, we should think about what to do with these - I'd appreciate your input to that! Marked as draft for now -- before the merge the second comments should either be dropped, or a follow up commit be made for whatever you imagine to happen with the `PECL` mentions.~~

### DIFF
--- a/skeletons/method.xml
+++ b/skeletons/method.xml
@@ -131,7 +131,7 @@ Procedural synopsis goes after OOP synopsis.
 <![CDATA[
 <?php
 if ($anexample === true) {
-    echo 'Use the PEAR Coding standards';
+    echo 'Use the PER Coding standards';
 }
 if ($thereisoutput === 'and it is multiple lines') {
     echo 'Use a screen like we did below';
--- a/skeletons/method.xml
+++ b/skeletons/method.xml
@@ -131,7 +131,7 @@ Procedural synopsis goes after OOP synopsis.
 <![CDATA[
 <?php
 if ($anexample === true) {
-    echo 'Use the PER Coding standards';
+    echo 'Use the latest PHP-FIG PER Coding style for PHP code examples: https://www.php-fig.org/per/coding-style/';
 }
 if ($thereisoutput === 'and it is multiple lines') {
     echo 'Use a screen like we did below';
--- a/skeletons/book.xml
+++ b/skeletons/book.xml
@@ -3,14 +3,12 @@
  <title>Long title of extension</title>
  <titleabbrev>Extname</titleabbrev>
 
- <!-- {{{ preface -->
  <preface xml:id="intro.extname">
   &reftitle.intro;
   <simpara>
    Extension introduction.
   </simpara>
  </preface>
- <!-- }}} -->
 
  &reference.extname.setup;
  &reference.extname.constants;
--- a/skeletons/classname.xml
+++ b/skeletons/classname.xml
@@ -5,19 +5,16 @@
  
  <partintro>
 
-  <!-- {{{ ClassName intro -->
   <section xml:id="classname.intro">
    &reftitle.intro;
    <simpara>
     Description of the class.
    </simpara>
   </section>
-  <!-- }}} -->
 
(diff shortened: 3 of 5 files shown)


skeletons/configure.xml
```diff
@@ -4,7 +4,7 @@
 
  <!-- If the extension doesn't need to be installed (built-in extension) use this entity -->
  <!-- &no.install; -->
- <!-- There are various PECL related entities in language-snippets.ent -->
+ <!-- There are various PECL related entities in language-snippets.ent / todo: decide PECL removal -->
```
kamil-tekiela: Don't add todos. Just ignore this comment for now if we are unsure if it can be deleted.

skeletons/method.xml
```diff
@@ -131,7 +131,7 @@ Procedural synopsis goes after OOP synopsis.
 <![CDATA[
 <?php
 if ($anexample === true) {
-    echo 'Use the PEAR Coding standards';
+    echo 'Use the PER Coding standards';
```
kamil-tekiela: What difference does this make?

skeletons/versions.xml
```diff
@@ -5,9 +5,9 @@
 <versions>
  <function name='foo' from='PHP 7 &gt;= 7.4.0, PHP 8'/>
  <function name='bar' from='PHP 8'/>
- <function name='baz' from='PECL extname &gt;= 0.4'/>
- <function name='quz' from='PECL extname 0.1-0.3' removed="PECL extname 0.3"/>
- <function name='classname::methodname' from='PECL extname &gt;= 1.1'/>
+ <function name='baz' from='PECL extname &gt;= 0.4'/><!--  / todo: decide PECL removal -->
```
kamil-tekiela: No, I think they need to stay.

skeletons/configure.xml
NickSdot: Please see the PR description; I don't plan to get the second commit merged -- it's to point y'all to what I'd appreciated feedback/direction to.

kamil-tekiela:
If we decide that we no longer allow PECL in the manual then the skeletons can be updated. I don't know who is using the skeletons and for what, but as of now, PECL extensions are still documented in the PHP manual. Either way it's a distinct change from removal of the editor fence comments.

skeletons/method.xml
NickSdot: > What difference does this make?

Defined here as new standard to follow in docs: https://github.com/php/doc-base/blob/master/docs/cs-for-examples.md#requirements

NickSdot:
~~> but as of now, PECL extensions are still documented~~

~~But no new docs added. So why would the skeleton/templates keep it?~~


**Edit:** 
> If we decide that we no longer allow PECL in the manual then the skeletons can be updated. I don't know who is using the skeletons and for what, but as of now, PECL extensions are still documented in the PHP manual. Either way it's a distinct change from removal of the editor fence comments.

You are right, having this discussion mixed in here doesn't make sense. I opened https://github.com/php/doc-base/pull/310

skeletons/method.xml
NickSdot: Pushed a commit to make it more clear.

kamil-tekiela (approved):
Ok. But I don't know know if we should have these skeletons.

NickSdot:
> Ok. But I don't know know if we should have these skeletons. 

Personally I found them helpful to get an idea for what is available without checking individual docs pages -- as in having a full overview at a glance.

jordikroon:
It's good to have skeletons _somewhere_ as they should reflect the recommended latest style guide and it helps new contributors to get started. They are not a burden to maintain either.

jordikroon (approved):

PR #311, opened 2026-06-27
Creates the do-not-translate mark

alfsb:
Updates the revcheck libraries to detect the `<?do-not-translate?>` marker in tracked files, allowing files to automatically declare that they are untranslatable and/or should not be copied into translations.

The `doc-base/scripts/broken.php` file was also modified along the same lines.

After this is merged, and DNT files are marked as so on `doc-en`, a lot of exceptional code can be removed from revcheck libs. This aligns in making `doc-base` more general, as it becomes less dependent on filename conventions, to facilitate the eventual splitting of external extensions manual from the language manual.

Reviews and comments welcome.

As this is a far reaching feature, plan to wait at least for two weeks before merging it,

### DIFF
--- a/docs/editing.md
+++ b/docs/editing.md
@@ -38,6 +38,9 @@ Changes in the English version are eventually picked up by the translators.
 If a change doesn't affect translations (e.g. fixing a typo in English) then the
 commit message should start with `[skip-revcheck]`.
 
+Files that are not to be translated, or copied on translations, should include
+an `<?do-not-translate?>` mark.
+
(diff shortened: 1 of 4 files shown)


Crell (approved):

alfsb:
Merged, tested, No functiional changes detected. Now it is possible to .xml files to mark thenselfs as to be ignored by translations and revcheck. Next is the marking of the actual files, and finally to remove the name conventions for exclusions from revcheck.

PR #314, opened 2026-07-04
Prevent doc-lang entities loading when doc-en file removed

jordikroon:
Prevents https://www.php.net/manual/en/install.windows.tools.php from rendering in Chinese.

### DIFF
--- a/scripts/file-entities.php
+++ b/scripts/file-entities.php
@@ -137,13 +137,14 @@ function pushEntity( string $name , string $text = '' , string $path = '' )
     global $entities;
     global $mixedCase;
     global $englishEntities;
+    global $lang;
 
     $name = str_replace( '_' , '-' , $name );
     $path = str_replace( '\\' , '/' , $path );
 
     // Prevents doc-en to load entities from doc-lang. Resulting in it rendering
     // a different language when the underlying doc-en XML file has been removed.
-    if ( $englishEntities !== null && ! isset( $englishEntities[ $name ] ) )
+    if ( $lang === 'en' && $englishEntities !== null && ! isset( $englishEntities[ $name ] ) )
     {
         $text = ' ';
         $path = '';
--- a/scripts/file-entities.php
+++ b/scripts/file-entities.php
@@ -87,14 +87,16 @@ echo "Running file-entities.php... ";
 
 $entities = [];
 $mixedCase = [];
+$englishEntities = null;
 
 generate_file_entities( $root , "en" );
 generate_list_entities( $root , "en" );
 
-if ( $lang != "" )
-    generate_file_entities( $root , $lang );
-
 pushEntity( "global.function-index", path: realpain( __DIR__ . "/.." ) . "/funcindex.xml" );
+
+$englishEntities = $entities;
+if ( $lang && $lang !== "en" )
+    generate_file_entities( $root , $lang );
 
 if ( ! $chmonly )
     foreach( $entities as $ent )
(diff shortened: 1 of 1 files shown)


alfsb:
This is a difficult one. The change is simple enough, but the consequences are not.

If the URL resolution of that page is wrong (in chasing the page in another language instead of 404 error), then both URL resolution and here should be changed in principle. But if URL resolution is changed (to what behaviour, other than the actual one?), then the change there may be moot. But the consequences are not.

The main point is that this change immediately break all translations, This change will avoid the creation of `&install.windows.tools;` entity, that is used in all [translations](https://github.com/php/doc-de/blob/08e5b7a4e17be6c5aceb2563d4c3c5d482205c56/install/windows/index.xml#L31), breaking then all.

This is the direct opposite of changes like https://github.com/php/doc-base/pull/303 , that maximize building, even for invalid XML, instead of this change that maximize breaking.

So, first, let's clarify what the final result of *URL resolution* should be. If URL resolution hits a non existent file, should it:
1. 404 always;
2. presents the file in another language, in all cases;
3. presents the file in another language, except for english;
4. other behaviour?

The answers delimit where the changes are to be made, and in what direction. In fact, one change I did on https://github.com/php/doc-base/pull/313 is to *keep* the actual behaviour, as to maximize the building of all translations.

Another way to pose this question is this: should all translations break, because URL resolution is not producing some specific result, for english or in the general case?

alfsb (commented):
As commented on Discord, I'm slightly opposed to this change.

I’d much prefer to change web-php/error.php into the direct, desired behaviour, than to change the result indirectly, by breaking all translations each time a file is removed (that will happen *a lot*, now the manual will be splited).

But let's have it open for a while, for further comments.

jordikroon:
So policy wise the docs should always fall back to doc-en when a translation for it doesn't exist. doc-en should be considered the main language. Thus if a file doesn't exist it should not translate to the one that comes after it. Instead it should return a 404 instead. If instead we want a redirect in place for it, that should be optional I believe.

Technically said the build didn't fail. Instead it creates an empty entity instead. You can see it in this PR as well, if this would break it the tests would also fail, right? 
`<!ENTITY install.windows.tools ' '>`

However it did result in having an empty entity reference for all translations. Which was an oversight. That should be covered with a check for $lang === "en"

So in short:
 - /en/install.windows.tools.php should give a 404
 - /zh/install.windows.tools.php should resolve

The other way round. If //zh/install.windows.tools.php wouldn't exist but /en/ would, it should fall back to /en/.

alfsb:
> * /en/install.windows.tools.php should give a 404

Then `web-doc/error.php` needs to be changed to not switch to other languages from `en`.

> * /zh/install.windows.tools.php should resolve

If by resolve this means to render the chinese page, at least for a while, then the `&install.windows.tools;` should be created and not empty, at least for `doc-zh`.

For this case, file entity creation should be maximal, as in the file entities are the union of all entities of files of languages in consideration (unique language, or fallback+translation), instead of the intersection of all files of the first language, of languages in consideration.

> The other way round. If //zh/install.windows.tools.php wouldn't exist but /en/ would, it should fall back to /en/.

Covered in point one. There is a switch on 404, but en/ is excepcional.

> Technically said the build didn't fail. Instead it creates an empty entity instead. You can see it in this PR as well, if this would break it the tests would also fail, right?

Creating a blank file entity will avoid the breaking of translations by the missing entity, but will erase the file's text of all translations immediately. If the "soft erased" file contains IDs or targets of XIncludes, this will start generating ID failures and manual parts will be erased immediatally, before translations can inspect the damage.

alfsb:
To make the example a little more concrete, consider that of all 11 languages of manual, 10 are translations, where 2 of them are in bad shape, but the other 8 are in good shape, being fully translated, or partially translated and where the outdated files are not very old. Then consider a change like [this one](https://github.com/php/doc-en/commit/71e7f31df31da40b847201a1585836d2c72014b8) that moved text around.

After a change like this, is expected for all translations:

* to immediately break;
* to immediately erase manual parts;
* No change.

These options maps to:

* do not create file entities for extra files in translations;
* create empty file entities for extra files in translations;
* create normal file entities for extra files in translations;

I'm in favor of the last alternative. Creating file entities for files that exist only in translations may seem wrong at first, but is the lesser of evils, for translations in good shape.

PR #317, opened 2026-07-07
Automatically generated files must self-declare

alfsb:
It is a small change, but I will wait at least two weeks for reviews and comments, as this is not a file I know well.

### DIFF
--- a/scripts/gen-phpdoc-tz-list.php
+++ b/scripts/gen-phpdoc-tz-list.php
@@ -31,10 +31,14 @@
 	$others = $groupedList['Others'];
 	unset($groupedList['Others']);
 	$groupedList['Others'] = $others;
-?>
-<?php echo '<?'; ?>xml version="1.0" encoding="UTF-8"?>
-<!-- AUTO GENERATED, DO NOT TRANSLATE OR MODIFY BY HAND -->
 
+    // Some IDEs are very unhappy about having XML processing
+    // instructions *outside* PHP processing instructions.
+
+    echo "<?xml version='1.0' encoding='UTF-8'?>\n";
+    echo "<?do-not-translate?>\n\n";
+?>
+<!-- AUTO GENERATED, DO NOT TRANSLATE OR MODIFY BY HAND -->
 <!-- The script to generate this file is scripts/gen-phpdoc-tz-list.php -->
 <!-- in the doc-base repo. However, it should only need to be run when  -->
 <!-- the tzdb file is updated by the maintainer of DateTime/timelib.    -->
@@ -54,7 +58,7 @@
 <?php
     }
 
-    foreach ($groupedList as $group => $zones) { 
+    foreach ($groupedList as $group => $zones) {
         $groupColumns = $columns;
         $m = count($zones) > ($groupColumns-1) ? $groupColumns : count($zones); ?>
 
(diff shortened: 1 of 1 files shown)


scripts/gen-phpdoc-tz-list.php
```diff
@@ -31,10 +31,14 @@
 	$others = $groupedList['Others'];
 	unset($groupedList['Others']);
 	$groupedList['Others'] = $others;
+
+    // Some IDEs are very unhappy about having XML processing
+    // instructions *outside* PHP processing instructions.
```
kamil-tekiela: What exactly does that mean? Which line does this comment refer to and what should future developers be aware of?

alfsb: These lines refer to the echo immediately below. PHP itself processes ok files like `<php ?><xml ?>text<?php ?>`.

But the deleted line that contains `<?php echo '<?'; ?>` shows that is not the first time someone trips on this, so I place the code that "fails" in echos, to be easier to read instead of the deleted part.

alfsb: Without this, many linters, and Code IDE in particular, will permanently mark this file as an erroneous one.

kamil-tekiela:
```
    echo <<<'XML'
    <?xml version='1.0' encoding='UTF-8'?>
    <?do-not-translate?>

    <!-- AUTO GENERATED, DO NOT TRANSLATE OR MODIFY BY HAND -->
    <!-- The script to generate this file is scripts/gen-phpdoc-tz-list.php -->
    <!-- in the doc-base repo. However, it should only need to be run when  -->
    <!-- the tzdb file is updated by the maintainer of DateTime/timelib.    -->

    <appendix xml:id="timezones" xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink">
     &date.timezone.intro.title;
     &date.timezone.intro;
    
    XML;

    if (function_exists('timezone_version_get')) {
?>
```
Solves the problem and no need for the comment. It also gets rid of the other PHP end/start tag.

kamil-tekiela:
I believe the title of this PR should have been "Automatically generated files must self-declare." just like the commit. Is "thenselfs" a typo?

scripts/gen-phpdoc-tz-list.php
alfsb: Nice. I will try that.

alfsb:
> I believe the title of this PR should have been "Automatically generated files must self-declare." just like the commit. Is "thenselfs" a typo?

English is not my first language. Tried to be fancy, but failed. Changing.

scripts/gen-phpdoc-tz-list.php
alfsb: Worked! At least on my IDE, the file is not marked erroneous anymore. Also replaced back single to double quotes, as the original files does.

kamil-tekiela (approved):

jordikroon (approved):

jordikroon:
This looks good to me. Fine if you merge it now imho. It's kinda straightforward.

alfsb:
I will go for it, merging today. There are four wide projects in the works (8.5 changes, manual split, entity conversion, xi:fallback), and having as few PRs open as possible is an objective in itself.

PR #318, opened 2026-07-08
configure.php: prepare PhD auxiliary-file generation for move to php/phd

lacatoire:
Prepares moving PhD auxiliary-file generation (version.xml, sources.xml, fileModHistory.php) out to php/phd, without removing anything yet so this can land independently.

- Add `phd_conf_json()`, which serializes the build state into `doc-base/temp/phd-conf.json` as a hand-off consumed by php/phd.
- Fix the `--history-file` option, which set `SOURCES_FILE` instead of `HISTORY_FILE`.
- Keep the existing `phd_acronym()`/`php_history()`/`phd_sources()`/`phd_version()` functions for now; their removal (and the matching .gitignore cleanup) depends on php/phd#261.

Supersedes the previous full-removal approach on this branch.


alfsb:
Wow! That is exactly what I was thinking of. Paint me very impressed.

The moved files are mentioned in `.gitignore`, so they can also be removed here. My long term plan is to shrink this file to only `temp/`.

alfsb:
I think the doc-base side is ok. You can merge here at your earliest convenience, and yet, let this open a few days, for comments and/or reviews. I will do a heads up no Discord, to be sure.

alfsb:
If the PhD side goes without indications of reviews or merging, then I will change this side, keep the hand-off file and PhD functions and merge in one week's time. These are some performance and doc-ext changes that are easier to do without worrying of git conflicts risk.

alfsb:
As PhD side shows no sign of being merged, I tried to checkout, rebase and push directly into this PR, but failed. @lacatoire , please consider rebasing and hard reseting this into master HEAD, and then applying the changes of https://github.com/alfsb/doc-base/tree/pr318.

alfsb:
Thanks for the update. Merged it as is, so as to avoid more conflicts in future, and to allow the PhD side to be developed independently.

PR #319, opened 2026-07-09
Remove filename conventions for do-not-translate in revcheck

alfsb:
This is the final part of removing filename conventions for do-not-translate files in revcheck lib.

No big changes here, but this may make third_party_ext_documentation a little less surprising, and easier to debug. Depends on doc-en 5662. Plan to merge it by next friday, if output of revchecks of all languages show no differences.


scripts/translation/lib/RevcheckIgnore.php
```diff
@@ -33,30 +33,6 @@ public static function ignore( $filename ) : bool
...
-        if ( $filename == "appendices/extensions.xml" )
-            return true;
-        if ( $filename == "appendices/reserved.constants.xml" )
-            return true;
-        if ( $filename == "reference/datetime/timezones.xml" )
-            return true;
-
         // Only in translations
 
         if ( $filename == "translation.xml" )
```
kamil-tekiela, line 32: What is this file? Why is it ignored here?

scripts/translation/lib/RevcheckIgnore.php
kamil-tekiela: I really hate code comments. If you could remove the comments from this function while you are at it, it would be great.

scripts/translation/lib/RevcheckIgnore.php
alfsb: `translation.xml` exists only in translations, and serves to feed some static data to revcheck, both local file and web hosted versions. Ironically, revcheck itself marks this file as an extraneous one, so it needs an exception.

`<?do-not-translate?>` wont work on files outside doc-en. In fact, any do-not-translate file in translations is an indication of a copy error.

And is placed here, as all exceptions being centralized in one place, instead of being tucked away somewhere where it can be easily forgotten.

kamil-tekiela: Well ok. I am not very familiar with how translators work. It's strange that they execute revcheck on a translation and not doc-en.

alfsb: Running revcheck *in* doc-en will result in... nothing, as the nothing to compare of. Revcheck runs in translations, so comparisons are done against doc-en.

kamil-tekiela: I meant that it's strange that we need to skip the file that is not present on doc-en.

alfsb: This file contains per language translations summaries and names/nicks/contacts. If this file is to be moved to doc-en, all translators will need write access on doc-en in principle. For now, this file exists only on translations.

Revcheck generates TODOs lines about XML files that exist only on translations, so they can be removed there. Without this exception, no revcheck would be ever empty/complete. And removing this XML file will break revcheck. So it would be an eternal pending item otherwise.

kamil-tekiela (approved):

alfsb:
Post tested. Detected non-expected differences on revcheck output.

Some "do not translate" files that exist on translation are marked as outdated, because revcheck cannot express the concept of files "do not translate, only in doc-en". TODOed a change in revcheck lib and script to create a new section.

PR #321, opened 2026-07-14
Trim the XML output file

alfsb:
This PR trims the generated Docbook XML file of PHP Manual, assembled by doc-base/configure.php. It strips:

1. The DOCTYPE node (after entity expansion);
2. Duplicated XML namespaces;
3. XML comments;
4. Insignificant whitespaces of Docbook.

Statistics:

- File reduction, bytes: -30,90%, from 44736907 to 30913772;
- File reduction, lines: -68,04%, from 1404683 to 448958;
- Maximum resident set size: -16,31%, from 1781788 to 1491252 kb.
- Time increase: about +0.25s

I'm surprised that the time increase is so small, but more surprised to see a reduction in Maximum resident set size. This means that the peak memory occurs later in configure (or any of its subs scripts), after the initial assembly.

Testing with PhD will be done in future. I expect no functional changes, but a full test is necessary anyways. PhD developers may need to check this very critically. Included a tool to recalculate the list, in the case the RNGs are updated.

Comments and reviews are welcome. Plan to wait at least two weeks before merging.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -208,7 +208,7 @@ function print_dom_errors()
     }
 }
 
-function print_xml_errors( bool $omitNoFile = false )
+function print_xml_errors()
 {
     global $ac;
     $report = $ac['LANG'] == 'en' || $ac['XPOINTER_REPORTING'] == 'yes';
@@ -228,7 +228,7 @@ function print_xml_errors( bool $omitNoFile = false )
         $line = $error->line;
         $clmn = $error->column;
 
-        if ( $file == '' && $omitNoFile )
+        if ( $file == '' )
             continue;
 
         if ( str_starts_with( $mssg , 'XPointer evaluation failed:' ) && ! $report )
@@ -756,7 +756,7 @@ function dom_load( DOMDocument $dom , string $filename , bool $firstLoad ) : boo
     $ret = $dom->load( $filename , $options );
     if ( $ret && $firstLoad )
     {
-        print_xml_errors( true );
+        print_xml_errors();
         xml_trim_first( $dom );
     }
     return $ret;
--- a/configure.php
+++ b/configure.php
@@ -208,11 +208,10 @@ function print_dom_errors()
     }
 }
 
-function print_xml_errors()
+function print_xml_errors( bool $omitNoFile = false )
 {
     global $ac;
     $report = $ac['LANG'] == 'en' || $ac['XPOINTER_REPORTING'] == 'yes';
-    $output = ( $ac['STDERR_TO_STDOUT'] == 'yes' ) ? STDOUT : STDERR ;
 
     $errors = libxml_get_errors();
     libxml_clear_errors();
@@ -220,9 +219,7 @@ function print_xml_errors()
     $filePrefix = "file:///";
     $tempPrefix = realpath( __DIR__ . "/temp" ) . "/";
     $rootPrefix = realpath( __DIR__ . "/.." ) . "/";
-
-    if ( count( $errors ) > 0 )
-        fprintf( $output , "\n" );
+    $firstBreak = false;
 
     foreach( $errors as $error )
     {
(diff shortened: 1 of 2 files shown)


alfsb:
Got curious, and tested each reduction.
```
Final size  Component   Reduction   Red%

44736907	original		
42815943	DTD	        -1920964	-4,29%
39039498	xmlns	    -3776445	-8,44%
33554966	comments	-5484532	-12,26%
30913772	dbwsi	    -2641194	-5,90%
```

docbook/docbookwsi.php
```diff
@@ -0,0 +1,193 @@
...
+XML document that follows the RelaxNG specification.
+
+If run with a second XML argument, the script will calculate all savings
+that can be done by stripping these insignificant whitespace between
+elements.
+
+See mentions of 'docbookwsi' on source code for parts that need to be
+updated if/when a new version of Docbook is used.                */
+
+$arg0 = array_shift( $argv ) ?? null;
```
jordikroon: This variable should be `$argv0`, print below will give `Undefined variable $argv0`.

docbook/docbookwsi.php
```diff
@@ -0,0 +1,193 @@
...
+$rngFile = array_shift( $argv ) ?? null;
+$xmlFile = array_shift( $argv ) ?? null;
+
+if ( $rngFile == null )
+{
+    print "Usage: '$argv0' rngFile [xmlFile]\n\n";
```
jordikroon, line 24: What parameters did you use to test this and determine the list? Perhaps we should document that.

I tried it using this, and it seems to work fine.
 > php docbook/docbookwsi.php docbook/docbook-5.2.1/rng/docbook.rng .manual.xml

configure.php
```diff
@@ -666,28 +698,88 @@ function dom_saveload( DOMDocument $dom , string $filename = "" ) : string
...
+        $node = $comments[ $idx ];
+        $node->parentNode->removeChild( $node );
+    }
+
+    // See docbook/docbookwsi.php.
+
+    // These first five elements account for almost 50%
+    // of all insignificant Docbook whitespace in manual.
+
+    $trimList[] = 'refsect1';
```
jordikroon: I prefer to use this form given it's a static array. But that's preference.

```php
$trimList = [
    // These first five elements account for almost 50%
    // of all insignificant Docbook whitespace in manual.
    'refsect1',
    'varlistentry',
    '...',
];
```

alfsb:
I'm delaying the merging of this PR, for now. There are some warnings being silenced on configure.php, and it is easier to check for insignificant differences without a change like this.

alfsb:
Rebased. Trim doctype, comments and duplicated namespaces. Removed trim of inter element whitespaces for now, as to avoid surfacing new PhD render errors.

Final reductions:
- Time: 11.15 sec to 9.80 sec;
- Peak resident set: 1.80 Gb to 1,50 Gb;
- Lines: 1.45 M to 1.17 M.

Also, exports DOCTYPE node as doc-base/temp/doctype.dtd, so code that needs to know all DTD entities used in manuals can consult this file.

Comments and reviews are welcome. Plan to merge this in one week.

alfsb:
> Also, exports DOCTYPE node as doc-base/temp/doctype.dtd, so code that needs to know all DTD entities used in manuals can consult this file.

@jordikroon , this may interest you. This has the potential to eliminate all `<entities>` loading for docbook-cs, because AFAIK this node contains all entities used in a assembled manual, in correct order, so docbook-cs can consume only this file, and parse only one format, to get all entities of manual at once.

My next step in configure is to minimize the DOCTYPE node to a single configure definition and usage, and by so, to restrict the SGML/DTD entity subsystem to only this line, in all manual.xml.

alfsb:
Delaying this again, as to wait for DTD minimization. DTD minimization on doc-base side produces identical XML, but on doc-en side produces XML differences, including in namespaces.

Let them run and render for a few days before making a change that eliminates about 85k of XML namespace (re)declarations.

alfsb:
Tested today, only in doc-en for now. Detected some surprising differences, documented in https://github.com/php/phd/issues/265 . Nothing that blocks this into being merged, but surprising differences anyway.

I plan to wait another full week, before merging this. Further comments and reviews are welcome.

alfsb:
Merged. `docbook/docbookwsi.php` is added, but unused. PhD is already [surfacing](https://github.com/php/phd/issues/265) HTML differences for insignificant XML changes (doctype, comments, duplicated namespaces), with also *file differences*. Let's wait until these are worked out before adding more differences caused by insignificant Docbook whitespace removal.

PR #322, opened 2026-07-16
Use a language directory's manual.xml when present

lacatoire:
When a language directory contains its own `manual.xml`, `configure.php` now builds from it instead of the main manual skeleton in doc-base. When absent, the doc-base `manual.xml` is used as before, so all existing builds (en and translations) are byte-for-byte unaffected.

This lets a manual be fully self-contained in its own repository: it can declare its own entity wiring and reference only its own books. Motivation is the third-party extension manual split ([lacatoire/doc-contrib](https://github.com/lacatoire/doc-contrib)), where reusing the main skeleton fails because it references ~160 core books the contrib repo does not provide. Supersedes the file-entities scanning approach discussed in #320.

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -674,7 +674,19 @@ function dom_saveload( DOMDocument $dom , string $filename = "" ) : string
 echo "Creating monolithic temp/manual.xml... ";
 $dom = new DOMDocument();
 
-if ( dom_load( $dom , "{$ac['srcdir']}/{$ac["INPUT_FILENAME"]}" ) )
+// A language directory may ship its own manual.xml to build a self-contained
+// manual (e.g. the third-party extension manual) that references only its own
+// books, instead of the main manual skeleton in doc-base. When absent, the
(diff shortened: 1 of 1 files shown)


alfsb (approved):
This is ok for me, and aligns with what I'm developing in doc-en 5682. But it may conflict with doc-base 321, so a bit of manual merge would be necessary.

In the meantime, apply these changes locally, so you are not blocked while all configure.php PRs are not merged.

alfsb:
Testing 321 will take longer that I expected. To not block the advance of third-party docs, this PR can be merged at any time, even more so if the rendeding and hosting advances.

alfsb:
Please test the PR 324, for a version that minimizes relative paths problems, and is also prepared for disappearance of DTD entity files (at least on doc-en side).

Also, consider using the DOCTYPE node of doc-en 5682 in the doc-ext manual.

Please comunique any problem found, and if everything goes well, consider closing this PR, in favor of the other two.

PR #325, opened 2026-07-21
Remove acronyms entity file from doc-base

jordikroon:
Suggested by @alfsb

Moving Acronyms from doc-base to doc-en. This allows Acronyms to be translated, as well as to more easily phase out unused acronyms.

This is a set of 3 PRs
doc-en: https://github.com/php/doc-en/pull/5695
doc-base: https://github.com/php/doc-base/pull/325 (you are here)
phd: https://github.com/php/phd/pull/262


lacatoire:
Nice move. Acronym expansions were one of the last reader facing strings stuck in English, and this puts them in the right place.

alfsb:
This may conflict with PR 381, but as noted [there](https://github.com/php/doc-base/pull/318#issuecomment-4936986047), I may change PR 381 as the hand off is done first, and deletion of PhD functions is done after PhD side is validated and merged.

In fact, I ask for your testing of PhD side of PR 381. Comment there if there is any change of hand off file contents or format, as to facilitate this process.

jordikroon:
Currently tests fail because doc-base now depends on doc-en. Once that's merged the entities should resolve for both doc-en and other languages.

This might conflict with 381, but if it does, it would be resolved fairly quickly as it's a 3 line removal that is covered in both this PR as well as in 381. Will give it a review tomorrow.

alfsb (approved):
Tests now pass, as XML entities landed on doc-en.

alfsb:
With the exception of changes in configure.php, I think this can be merged. Very few translations picked the change that would not be visible anyways. without the PhD side, but there was enough test to see if doc-en and translations behaves ok with the new entities hosted on doc-en.

Tomorrow I will start working into further shrinking global.ent and lowering IDREFs errors into warnings in translations, so they can be merged soon after manual split settles down.

jordikroon:
The commit containing the configure.php changes has been dropped from this PR. So it shouldn't cause any merge conflicts. @alfsb

PR #326, opened 2026-07-23
configure.php: recover failed XIncludes inside classsynopsis

lacatoire:
`xinclude_residual_fixup()` recovers failed XIncludes on translations per
parent element, but had no `classsynopsis` case, so a failed include targeting
a `methodsynopsis` or `fieldsynopsis` inside a class synopsis hard-failed the
build (`Unknown parent of failed XInclude: classsynopsis`).

Adds a `classsynopsis` case injecting a `<classsynopsisinfo role="comment">`
marker, a valid child per the DocBook 5.2 RNG.

Related:
- php/doc-en#5699 removes empty `<xi:fallback/>` that masked such failures
- php/doc-en#5700 the Imagick fix whose failure first hit this path

@alfsb

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -839,6 +839,9 @@ function xinclude_residual_fixup( DOMDocument $dom )
             case "variablelist":
                 $fixup = "<varlistentry><term></term><listitem><simpara>$alert</simpara></listitem></varlistentry>";
                 break;
+            case "classsynopsis":
+                $fixup = "<classsynopsisinfo role=\"comment\">$alert</classsynopsisinfo>";
+                break;
             default:
                 echo "  (Unknown parent of failed XInclude: $parent)\n";
                 $hardfail = true;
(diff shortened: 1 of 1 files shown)


alfsb:
Got curious about the `role="comment"` part. I tested without the attribute (and after removing all `<xi:fallback/>locally), and it passes RNG 5.2.1 validation, on PHP 8.1.2-1ubuntu2.25, libxml 2.9.13. Checked the output to be sure that it was generated.

Does it break something locally, or does the missing attribute cause omitions/problems on the PhD side?

I prefer to keep fixups as minimal as possible. so no attribute or `role='comment'` (without the escape) is preferable. That said, it is a very ok change, that I may merge immediately.

alfsb (approved):

lacatoire:
Good question. Dropping the attribute passes RNG, but it changes the PhD rendering. In `format_classsynopsisinfo` (Generic/XHTML.php:1145), a role-less `classsynopsisinfo` is treated as the class-body opener: it appends  `{`, sets `classsynopsis` close=true and asserts legacy rendering. With the `comment` role the marker renders as `/* ... */`, which is what the fixup wants.

I've switched to single-quoted role='comment' to drop the escaping, as you suggested.

alfsb:
Nice. I really do not know these PhD specific behaviors.

Plan to merge this by monday, before any xi:fallback deletions.

jordikroon (approved):

Issue #327, opened 2026-07-23
Why doc-it is not online?


DavidePastore:
I saw @jordikroon said in [another issue](https://github.com/php/doc-base/issues/278#issuecomment-4816480080):

> The documentation is rebuilt every 2 hours. However as for the Italian translation, it's currently broken because it's not fully up-to-date combined with a migration to a new server. Once the translation passes the build, it should re-appear.

But it's not clear to me why doc-it is not online. Could you please tell me? Thanks!

jordikroon:
Hi @DavidePastore

Thank you for once and again reporting this.

doc-it is offline because the Italian translation currently fails XML validation during the server build, so no output gets published. This is what the build stops on right now (from a CI run in #328):

```
Run php doc-base/configure.php --disable-libxml-check --enable-xml-details --redirect-stderr-to-stdout --with-jing=no --with-lang=it
configure.php on PHP 8.3.6, libxml 2.9.14

Checking for source directory... /home/runner/work/doc-base/doc-base/doc-base
Checking for output filename... /home/runner/work/doc-base/doc-base/doc-base/.manual.xml
Checking whether to include CHM... no
Checking for PHP executable... /usr/bin/php8.3
Checking for language to build... it
Checking whether the language is supported... yes
Checking for partial build... no
Checking whether to enable detailed XML error messages... yes

doc-base: ab6b7401a113899a73e22741c2e002f2221aae8f
en:       462dddda87baa86fd760dcb1e3a0d2096e00a59b
it:       ec2787e087db5278a64f80914a0fbad3ab23f1d5

Running file-entities.php... done: 12414 entities.
Running text-entities.php... (skiped it/entities) done: 139 entities, 135 untranslated, 1 errors.
(Run 'php /home/runner/work/doc-base/doc-base/doc-base/scripts/text-entities.php en it --debug' for details.)
Checking for if we should generate a simplified file... no
Creating monolithic temp/manual.xml...  done.
Expanding XIncludes... done: 1846 tags replaced.

Failed XIncludes, manual parts will be missing. Unresolved xpointers:
- bcmath-number.round..parameters.mode
Dumped file: temp/manual.err. Inspect residual xi:include tags in this file.

Validating temp/manual.xml (libxml)... failed.

[error manual.xml 1319020:0] IDREF attribute linkend references an unknown ID "ini.fatal-error-backtraces"
[error manual.xml 31189:0] IDREF attribute linkend references an unknown ID "language.types.void.casting"
[...]

PR #328, opened 2026-07-24
Add --with-jing option and exercise both validators in CI

jordikroon:
Server build process doesn't use `jing` for XML validation. Which is related to why doc-it https://github.com/php/doc-base/issues/327 is unable to build.

This is a first step towards fixing it by making sure we have a correct CI pipeline that build with and without `jing`.

CI is expected to fail for doc-it with the message below:

```
Failed XIncludes, manual parts will be missing. Unresolved xpointers:
- bcmath-number.round..parameters.mode
Dumped file: temp/manual.err. Inspect residual xi:include tags in this file.

Validating temp/manual.xml (libxml)... failed.

[error manual.xml 31350:0] IDREF attribute linkend references an unknown ID "language.types.void.casting"
[error manual.xml 31460:0] IDREF attribute linkend references an unknown ID "language.types.void.casting"
[error manual.xml 1319249:0] IDREF attribute linkend references an unknown ID "ini.fatal-error-backtraces"

Eyh man. No worries. Happ shittens. Try again after fixing the errors above.
```

### DIFF
--- a/configure.php
+++ b/configure.php
@@ -85,6 +85,8 @@ Package-specific:
   --with-php=PATH                Path to php CLI executable [detect]
   --with-lang=LANG               Language to build [{$acd['LANG']}]
   --with-partial=my-xml-id       Root ID to build (e.g. <book xml:id="MY-ID">) [{$acd['PARTIAL']}]
+  --with-jing=auto|yes|no        Validate with Jing (requires Java) instead of
+                                 libxml [{$acd['JING']}]
   --disable-broken-file-listing  Do not ignore translated files in
                                  broken-files.txt
   --disable-xpointer-reporting   Do not show XInclude/XPointer failures. Only effective
@@ -294,6 +296,7 @@ $acd = array( // {{{
     'TRANSLATION_ONLY_INCL_BEGIN' => '',
     'TRANSLATION_ONLY_INCL_END' => '',
     'XPOINTER_REPORTING' => 'yes',
+    'JING' => 'auto',
 ); // }}}
 
 $ac = $acd;
@@ -423,6 +426,14 @@ foreach ($_SERVER['argv'] as $k => $opt) { // {{{
 
         case 'xpointer-reporting':
             $ac['XPOINTER_REPORTING'] = $v;
+            break;
+
+        case 'jing':
+            if (!in_array($v, ['auto', 'yes', 'no'], true)) {
+                errbox("Invalid value for --with-jing: $v (expected auto, yes or no)");
+                errors_are_bad(1);
+            }
+            $ac['JING'] = $v;
             break;
 
         case '':
@@ -989,11 +1000,15 @@ function xml_validate( $dom )
 
     // Jing is faster, but depends on Java.
 
-    $out = null;
-    $ret = null;
-    exec( "java -version 2>&1" , $out , $ret );
+    $jing = $GLOBALS['ac']['JING'];
 
-    if ( $ret == 0 )
+    if ( $jing === 'auto' )
+    {
+        exec( "java -version 2>&1" , $out , $ret );
+        $jing = $ret === 0 ? 'yes' : 'no';
+    }
+
+    if ( $jing === 'yes' )
         xml_validate_jing();
     else
         xml_validate_libxml( $dom );
@@ -1006,16 +1021,11 @@ function xml_validate_jing()
 
     echo "Validating temp/manual.xml (jing)... ";
 
-    $out = null;
-    $ret = null;
     $schema = RNG_SCHEMA_FILE;
     $cmdJing = "java -Djdk.xml.totalEntitySizeLimit=300000 -jar {$srcdir}/docbook/jing.jar {$schema} {$idempath}";
     exec( $cmdJing , $out , $ret );
 
-    if ( ! is_array( $out ) )
-        $out = [];
-
-    if ( $ret == 0 )
+    if ( $ret === 0 )
     {
         echo "done.\n";
         return;
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -58,5 +58,8 @@ jobs:
           php doc-base/scripts/qa/extensions.xml.php --check
           php doc-base/scripts/qa/section-order.php
 
-      - name: "Build documentation for ${{ matrix.language }}"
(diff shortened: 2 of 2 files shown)


configure.php
```diff
@@ -425,6 +428,14 @@ function find_xml_files($path) // {{{
             $ac['XPOINTER_REPORTING'] = $v;
             break;
 
+        case 'jing':
+            if (!in_array($v, ['auto', 'yes', 'no'])) {
```
kamil-tekiela:
```suggestion
            if (!in_array($v, ['auto', 'yes', 'no'], true)) {
```

configure.php
```diff
@@ -989,11 +1000,17 @@ function xml_validate( $dom )
...
-    $out = null;
-    $ret = null;
-    exec( "java -version 2>&1" , $out , $ret );
+    $jing = $GLOBALS['ac']['JING'];
 
-    if ( $ret == 0 )
+    if ( $jing == 'auto' )
+    {
+        $out = null;
+        $ret = null;
```
kamil-tekiela: What are these for?

kamil-tekiela (changes requested):

configure.php
jordikroon: Left over from the old code. Likely there because the line after this uses it by reference. Which PHP already does for you. Removed ;-)

configure.php
```diff
@@ -1023,16 +1021,14 @@ function xml_validate_jing()
...
-    $ret = null;
     $schema = RNG_SCHEMA_FILE;
     $cmdJing = "java -Djdk.xml.totalEntitySizeLimit=300000 -jar {$srcdir}/docbook/jing.jar {$schema} {$idempath}";
     exec( $cmdJing , $out , $ret );
 
     if ( ! is_array( $out ) )
         $out = [];
 
-    if ( $ret == 0 )
+    if ( !$ret )
```
kamil-tekiela: === is better

was there ever a chance for the NULL value to survive?

jordikroon: It should be an int if we set $ret. So strict is fine here.

configure.php
```diff
@@ -1006,16 +1021,14 @@ function xml_validate_jing()
...
     echo "Validating temp/manual.xml (jing)... ";
 
-    $out = null;
-    $ret = null;
     $schema = RNG_SCHEMA_FILE;
     $cmdJing = "java -Djdk.xml.totalEntitySizeLimit=300000 -jar {$srcdir}/docbook/jing.jar {$schema} {$idempath}";
     exec( $cmdJing , $out , $ret );
 
     if ( ! is_array( $out ) )
         $out = [];
```
kamil-tekiela: Then the same applies here, right? This is dead code.

kamil-tekiela (approved):

configure.php
jordikroon: 100% obsolete code. I changed it, but honestly good to sweep through the whole file and modernise it at some point. Which is hopefully after we are done with all the major changes.

alfsb:
I think this is wrong, as calling configure.php --with-jing=yes on a machine where there is no Java, the code will blindly call xml_validate_jing(), that in turn `exec( "java -D ...")`, that fails (silently?).

IDREF errors are downgraded to warnings in translations, at validation step, but only in Java/Jing branches. See: https://github.com/php/doc-base/commit/f572f009bfd6113d5b9919f9eb49128ad7fd3bc4 .

The root problem can be more easily resolved by replicating the same downgrade on libxml branches, but only in translations. On doc-en, IDREFs should still be a hard failure. No `--with-jing` is necessary.

alfsb:
That is, to change `xml_validate_libxml()` as to clean libxml errors before the `$dom->relaxNGValidate( RNG_SCHEMA_FILE )`, and to inspect if the generated errors on failure.

if in_translation if error_count == ifref_count { echo done, return }

jordikroon:
> I think this is wrong, as calling configure.php --with-jing=yes on a machine where there is no Java, the code will blindly call xml_validate_jing(), that in turn `exec( "java -D ...")`, that fails (silently?).

I have added improved error handling for when Java is absent. So that should not fail silently anymore. Also Jing errors are handled better now with `2>&1` added to it.

> IDREF errors are downgraded to warnings in translations, at validation step, but only in Java/Jing branches. See: [f572f00](https://github.com/php/doc-base/commit/f572f009bfd6113d5b9919f9eb49128ad7fd3bc4) .
> 
> The root problem can be more easily resolved by replicating the same downgrade on libxml branches, but only in translations. On doc-en, IDREFs should still be a hard failure. No `--with-jing` is necessary.

This PR was not meant to be a fix for https://github.com/php/doc-base/issues/327, but rather a fix for the inconsistency between the production build, and the local or CI build. Given we allow both Jing and libxml validation, we should test both here. Otherwise it is bound to happen again one way or another.

So in other words, we need a follow up PR in order to fix doc-it and apparently doc-uk as well.

alfsb:
> I have added improved error handling for when Java is absent. So that should not fail silently anymore. Also Jing errors are handled better now with `2>&1` added to it.

I think it is now ok.

> So in other words, we need a follow up PR in order to fix doc-it and apparently doc-uk as well.

There are a few changes on text-entities, configure fixups, en revcheck arg passing. I will try to make those in one go, next week.

Also, any changes in doc-base will help the study of actions/checkout being stuck in the past.

alfsb:
On second thought, this option can be named something `--debug-ci` or `--all-validators`, and code changed to something like this
```
 if ( $hasJava || $debugCI )
    validateJing()
 if ( ! $hasJava || $debugCI )
    validateLibxml()
```
The reason is twofold. First, the configure is run really once in CI, instead of double run, almost doubling performance. Second, I do not see this option being used outside Github CI.

As its *raison d'être* is catching failure validation differences, both functions call exit, so either one failing will fail the entire (and unique) run.

alfsb:
In fact, something like, to also include the requirement that CI has to have Java:
```
 if ( $debugCI && ! $hasJava )
    // No Java error on CI machine
 if ( $hasJava || $debugCI )
    validateJing()
 if ( ! $hasJava || $debugCI )
    validateLibxml()
```

alfsb:
In fact, as I'm slowly making changes and documentating the GH CI behavirour, I may need to include the `--debug-ci` option anyways, as to expose verbosity of configure.php in this enviroment. So the merged option would something like:
```
  --debug-ci         Changes validation and verbosity for CI enviroments.
```

PR #330, opened 2026-07-25
Fix text-entities statistics, clean ups

alfsb:
Fix statistics on doc-base/text-entitites.php, that reported not translated files as other errors. Remove entities/*-dist files, and QoL fix in revcheck.

Comments and reviews wellcome. Plan to merge this in one week


alfsb:
Merged. Now the text-entities.php will count untranslated completely separated from other errors.

PR #331, opened 2026-07-26
Change manual configuration entity file extension from .inc to .ent

jordikroon:
This shouldn't cause any issues but it's required to fix docbook-cs to load LANG. 
Which is why https://github.com/php/doc-en/pull/5682/changes current reports:

```
manual.xml: XML parse error: Entity 'LANG' not defined
```

A .inc file is usually also parsed as a PHP file in many IDEs like PHPStorm, so changing it to a .ent is the best and easiest solution.


kamil-tekiela (approved):

alfsb:
Please, consider changing the file name to `conf.dtd`. I think this better matches the internal parameter entity used in manual.xml (`%configure;`) and file contents (Document Type Definition, dtd).

A patch I'm working on will do the suggested rename, and also will eliminate the `&LANG;` entity entirely. It is the only entity reference inside a XML attribute, and a sore thumb in general.

All .ent files in doc-base, and also in the root of doc-en are in fact .dtd files, not .ent files, but these are to be slowly transformed into .ent XMl files.

jordikroon:
I'm happy to do that change in a follow up PR. Before I make that change I will add compatibility to the docbook-cs, which should be a relative easy fix. 

With .inc I was fearful it might load PHP files by accident, so I wanted to get this out of the way. Expect a follow up in the next few days.

alfsb:
I did not see that it was merged, so my message arrived too late. doc-base 324 already does that change, and got conflicted. I will rebase it, so a follow up PR is not necessary.

PR #332, opened 2026-07-27
Alert in xi:include fixups, block/warning against functional xi:fallback

alfsb:
Add tofu (`U+FFFD � REPLACEMENT CHARACTER`) in empty elements to avoid trip PhD into generating invalid HTML, also block (on doc-en) and warning (in translations) in the presence of reachable functional `<xi:fallback>`s, that cause manual parts to go silently missing.

Comments and reviews are welcome. Plan to merge this after manual split.


alfsb:
The tofu thing is a workaround to the long standing https://github.com/php/phd/issues/247 and https://github.com/php/phd/issues/181 .

jordikroon:
Please use `&#xFFFD;` instead. It trips me over a little, and it's a special unicode character. Just like how `&amp;` is preferred usually.

This character is also not consistently rendered across browsers and/or operating systems, but for this purpose that's not a problem.

alfsb:
One alternative over tofu is to repeat the `$alert` in each inline element. Even more so now that it is "reduced", after I pushed the failed xpointer body into XML comment, that probably will not render.

That said, I prefer a single character that is not consistently rendered across browsers, than having random pages being consistently rendered wrong across browsers.

But the comment thing worries me. It will take a long time until I borrow a machine to test the PhD side, to see if this change in fact fixes both issues, *and* does not cause one immediate breakage.

alfsb:
Replaced tofy by alert.

Issue #199, opened 2024-12-04
Manual infra-estructure


alfsb:
Planed changes, big and small, that affect manual infra-estruture and languages.

#### Recheck

- [x] Single source revcheck (PR 111 174 176 180)
- [x] sequential `[skip-revcheck]` (PR 178)
- [ ] accidental `[skip-revcheck]` (PR 181)

#### Configure

- [X] Nested/recursive XIncludes (PR 194)
- [ ] XInclude failures less fatal on translations (PR 196)
- [ ] XInclude by `xml:id` (PR 198)
- [ ] Individual/bundled XML Entities (PR 183)
- [ ] Split `language-snippets.ent`
- [ ] Convert all other `.ent` files to normal XML

#### Idempotent build

- [ ] Create `doc-base/temp` and `.gitkeep` it
- [ ] Rewrite `file-entities.ent.in` to use `temp/`
- [ ] Recode `manual.xml.in` into (stable) `manual.xml`
- [ ] Move `.manual.xml` to `temp/` (coordinated)
- [ ] Move all remaining `.gitignore` entries to `temp/` (coordinated)

#### Performance/tools

- [ ] Extricate `qaxml` tools from lib revcheck
- [ ] Test and modify `qaxml` tools to be used in more languages
- [ ] R&D alternative for files entities

---

Not discussed, just a list of dreams and/or pet peeves

- Simplify git statuses on start of configure
- Stack trace like libxml errors
- Make `<?phpdoc?>` more general
- Remove all mentions of `<xi:fallback`>, XML and code
- Eliminate all manually edited `<!ENTITY>` files
- Mentoring translations about XInclude failures and `qaxml` alerts
- Move list of active languages to `doc-base` (coordinated) (PR 186)

PR #334, opened 2026-08-04
Add "Do not translate" and "XML Broken" sections to revcheck

lacatoire:
Both were already being computed and then dropped: files marked `<?do-not-translate?>` were detected and skipped without a trace, and libxml errors were collected by `XmlUtil` then cleared without ever being read. This lists them, in the file summary and in a section of their own.

Telling a real error apart from an unresolved entity was already solved in `scripts/broken.php`; that logic moves to a shared `XmlErrorFilter` so both callers follow the same rule. Only the entity files still holding entity declarations are skipped, as those DTD fragments never parse as standalone XML; the .ent files that already are XML are checked like any other.

Files marked do not translate are listed but kept out of the totals, so no translation sees its completion rate drop for files it is not expected to have. The `files` table gains an `xmlError` column, so the web report can say what is wrong and not only which file.

An empty file raised a `ValueError` in `XmlUtil::loadText()` and aborted the run. It is now reported as any other broken file.

Item "Do not translate and XML Broken sections in revcheck" of #199.

Web side, needed for the statuses to show up on doc.php.net: php/web-doc#68

Checked with php 8.4: `scripts/broken.php` output unchanged on doc-fr, doc-es, doc-en, doc-ru, doc-de, doc-it and doc-tr; the doc-fr report identical outside the new sections, totals included; XML detection agreeing with `broken.php` on the same six languages, with the same single finding.

### DIFF
--- a/scripts/translation/lib/RevtagParser.php
+++ b/scripts/translation/lib/RevtagParser.php
@@ -33,7 +33,22 @@ class RevtagParser
     static function parseDir( string $lang , RevcheckFileList $list )
     {
         foreach( $list->iterator() as $entry )
+        {
             $entry->revtag = RevtagParser::parseFile( $lang . '/' . $entry->file );
+
+            // Files are parsed here anyway, so reuse the errors already
+            // collected by XmlUtil, instead of loading everything again.
+            //
+            // Only .xml files are checked. Entity files are DTD fragments,
+            // never standalone XML, so they always fail to parse as such.
+
+            if ( str_ends_with( $entry->file , '.xml' ) == false )
+                continue;
+
+            $error = XmlUtil::$lastErrors[0] ?? null;
+            if ( $error != null )
+                $entry->xmlError = trim( $error->message ) . " [{$error->line},{$error->column}]";
+        }
     }
 
     public static function parseFile( string $filename ): RevtagInfo|null
--- a/scripts/translation/lib/RevtagParser.php
+++ b/scripts/translation/lib/RevtagParser.php
@@ -39,11 +39,18 @@ class RevtagParser
             // Files are parsed here anyway, so reuse the errors already
             // collected by XmlUtil, instead of loading everything again.
             //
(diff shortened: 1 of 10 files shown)


alfsb (approved):
This is impressive work. I will do some tests, and then merge in about a week. Further reviews and comments welcome.

The only comment is about the comment on revtag failing when XML is detected broken. This is true on empty files, but the (almost now extinct) real broken XML files can have valid revtags (as they are text searched) even if XML is misaligned or otherwise invalid.

scripts/translation/lib/RevtagParser.php
```diff
@@ -33,7 +33,22 @@ class RevtagParser
...
+        {
             $entry->revtag = RevtagParser::parseFile( $lang . '/' . $entry->file );
+
+            // Files are parsed here anyway, so reuse the errors already
+            // collected by XmlUtil, instead of loading everything again.
+            //
+            // Only .xml files are checked. Entity files are DTD fragments,
+            // never standalone XML, so they always fail to parse as such.
+
+            if ( str_ends_with( $entry->file , '.xml' ) == false )
```
alfsb, line 48: There are some .ent files that are in fact XML, and some .ent files that are DTD parts. I'm trying to slowly fix this, to have XML in .ent and .xml, and to move DTD entity files as .dtd, or erase them entirely, so in the end all .xml and .ent files will be XML.

Here, the test is about not adding errors for these .ent DTD files. Assume everything is XML in principle, and exclude the known case: files not .xml, that also contains `<!ENTITY`.
The file is "hot" in OS cache, so reloading these few not .xml cases here to check contents will cause no visible regressions.

alfsb:
Tested. Detected one change in doc-ru ("Up to date files" + 1), but otherwise ok. Plan to merge this in one week.

Further reviews and comments welcome.

alfsb (approved):
Difference in doc-ru is a change from `TranslatedOk` into `XmlBroken` transition, so it is expected.

PR #339, opened 2026-08-24
Do not mark do-not-translate files not translated

alfsb:
Files marked do-not-translated are showing in revcheck even if they are not present in translations. This is a small fix for that.

The full and correct fix, to detect and mark do-not-translate files present in translations, will need a wider change, as to avoid performance regressions.

Plan to merge this in one week, Reviews and commits welcome.


alfsb:
Updated with full do not translate support. Tested on all languages. Identical results, aside fixing files marked do-not-translate, not in translations, and yet showing in the revcheck list / database.

Some languages the time of calculations is reduced by half.

alfsb:
Merged, tested. All languages are ok. Most active languages converged to the correct status even with the misleading revcheck. Languages now can reach 100% again.

PR #340, opened 2026-09-04
Adapt broken.php for doc-extensions, and faster/simpler text-entities.php

alfsb:
Adapt broken.php to check for presence/absence of manual.xml, for detecting translations dirs, and also change re-quotation hell avoidance in text-entities.txt, making it more normal and slightly faster in translations.

I also considered changing revcheck to warn about the presence of `<xi:fallback>`s in translations, if not now, then after manual split is done, so mass changes can be done once.

Comments and reviews welcome, plan to merge this in one week.


alfsb:
Merged and tested. broken.php will now report xi:fallbacks still present in languages, like configure.php. There are quite a few maintained languages in this situation.

Next in line is reworking revlib a little, to generate a list of lines in TranslatedOk status (`doc-base/temp/files-ok.txt`), two counter files to externalize how many files are outdated more than one and two years in translations (`doc-base/temp/count-outdated-1y.txt` and `count-outdated-2y.txt`), and finally rewriting the logic in GitLogParser to set the parsing logic in more imperative form, instead of a push state machine, as it is now.

There are still two other projects that I'm working on, from time to time. The first one is a XML assembler in userland code, as to avoid the hard limits of libxml and also a form to resolve the long standing doc-base 72 (up to XInclude validation), and another is to make file-entities.php export file inclusion entities with a more uniform naming convention.

PR #342, opened 2026-09-09
Add check-structure.php, comparing translation structure at declared revision

lacatoire:
Compares the block structure of translated files with doc-en: the elements and their nesting, not the prose they contain.

Six translations already run it from their own copy, in three variants; one of them never received the fix that added `href` and `xpointer` to the compared attributes, and lets a translated XInclude target through.

Checked against doc-fr: 11359 files, 6 differences, the same ones its local copy reports.

First consumer: php/doc-fr#3544.


lacatoire:
Corrected PL build here https://github.com/php/doc-pl/pull/36

alfsb (approved):
Thanks for keeping the uncommon code style.

PR #344, opened 2026-09-16
List of TranslatedOk files and count of files outdated more than 1 and 2 years

alfsb:
This change makes revlib code to export a list of TranslatedOk files, so it will be possible to change quality and sync tools to observe only (or preferably) the files in this status.

Also exports counts of files outdated on translations, for more than 1 and 2 years, so struggling translations can be easily spotted and, for example, can receive a wide banner asking for help, and also abandoned translations can now be detected.

The statistics can be generated from the script on `doc-base/scripts/translations/configure.php`. I'm also seriously considering generating the same data on `doc-base/scripts/translations/genrevdb.php`, as this is the type of information that is already produced and used while building the manual in the main site.

For this to happen, is necessary to change a single line in `genrevdb.php`, from:

```php
$revcheck = new RevcheckRun( 'en' , $lang );
```

to

```php
$revcheck = new RevcheckRun( 'en' , $lang , true );
```


alfsb:
I was considering "completing" the function `xinclude_remove_fallback()`, into really removing any still present `xi:fallback` in translations, but decided against it, as it would not change the manual content in a meaningful way, and could also break various translations as result.

Comments and reviews wellcome. Plan to merge this in one week.

